PR #18, opened 2021-09-13
ENH: improve reading perfs for dmp files

### LINKED ISSUE #17: ENH: avoid combining struct + numpy
Currently `_io.dmp_io.read_chunks` uses a combination of `struct.unpack` and `numpy` to retrieve potentially large arrays, which is suboptimal as compared to using `numpy.fromfile` directly

neutrinoceros:
got a ~25% speedup for a ProjectionPlot example using a 660MiB dataset.
Only problem is that the reshaping is currently broken so the results are incorrect.
fix #17

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = yt_idefix
-version = 0.3.1
+version = 0.3.2
 description = An extension module for yt, adding a frontend for Idefix
 long_description = file: README.md
 long_description_content_type = text/markdown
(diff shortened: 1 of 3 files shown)


neutrinoceros:
fixed the bug. Now using the following benchmark:
```python
# benchmark.py
import yt
import yt.extensions.idefix
import time
import sys

def main(n:int=100) -> int:
    yt.set_log_level("error")
    start = time.monotonic()

    for _ in range(n):
        ds = yt.load("dump.0000.dmp")
        pp = yt.ProjectionPlot(ds, "z", ("gas", "density"))
    stop = time.monotonic()
    delta = stop - start
    print(f"Ran {n} loops in {delta:.1f} s")
    return 0

if __name__ == "__main__":
    sys.exit(main())
```
using a 660MiB dataset, I get
#### on `main`
```
Ran 100 loops in 262.4 s
```

#### this branch
```
Ran 100 loops in 197.9 s
```

PR #55, opened 2021-12-09
ENH: add explicit support for PLUTO vtk datasets

neutrinoceros:
Starting the process of specializing support for Pluto vtk files.

@xshaokun, it would be extremely helpful if you could do any of the following:
- provide an actual PLUTO vtk file I could integrate to the test suite (preferably small)
- validate that this patch works as intended for you. Namely, you should now be able to load Pluto vtk files without getting Idefix-specific warnings, and the dataset objets should now be of type `PlutoVtkDataset` instead of `IdefixVtkDataset`, eventhough this distinction doesn't currently change anything.


neutrinoceros:
Yes, Idefix's vtk format is based off Pluto's but it has unique stuff that is not present in Pluto. That's fine, there's already a way to pass the geometry value manually, so I'll assume it's sufficient for the time being.

xshaokun:
I want to test this version but encountered a problem.

```
In [3]: ds = yt_idefix.load("p2r003-trc/data.0000.vtk")
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-6560c72ba9af> in <module>
----> 1 ds = yt_idefix.load("p2r003-trc/data.0000.vtk")

AttributeError: module 'yt_idefix' has no attribute 'load'
```

I clone the git repo and add the path  into `$PYTHONPATH`. It looks like that the module level is messed up.

neutrinoceros:
> I clone the git repo and add the path into $PYTHONPATH. It looks like that the module level is messed up.

you shouldn't need to edit your $PYTHONPATH, and I'm not suprised that it doesn't work actually.
The way to install from source is to run `pip install -e .` from the top level of your clone.

neutrinoceros:
Anyway I think this is now ready for merging, and I was able to validate that it works correctly thanks to your data.
I'm going to release this as yt_idefix 0.8 now, feel free to open a new issue if needed !

neutrinoceros:
> So in this way, every time the package will be imported from source and no need to be installed again, right?

actually, no ! Because yt_idefix is currently pure-Python, it doesn't need to be recompiled, and the `-e` flag in `pip install -e .` means "editable", so you _don't_ need to reinstall each time you update the repo.
Note that I publish releases immediately after any relevant PR is merged, so you don't have to install from source to keep up with my updates, you also have the option of upgrading from PyPI with `pip install -U yt_idefix`.

xshaokun:
> the -e flag in pip install -e . means "editable", so you don't need to reinstall each time you update the repo.

That's exactly what I mean. Thanks! :)

neutrinoceros:
Oh sorry, my bad, I missed a negation when I first read your question.

PR #56, opened 2021-12-15
ENH: parse code unit, geometry and time (Pluto)

xshaokun:
Actually this is my first time to open a pull request in a pulic project, so I'm a little bit exited to have such an opportunity.

I have made the following changes:

- create a new method `_parse_header_file` for parsing header file `definitions.h`, which is searched from data directory or code directory (a new keyword argument `code_dir` is introduced, since pluto could specify a seperate location for output data.)
- In Pluto, code units can be defined through pre-defined constants. So I stored these contants in `definitions.py`.
- Imported unit values are stored in `self.parameters`

I tried to avoid modifying your existed code, hope that my coding is consistent with your convention.

TODO:

- [x] add new test data for units conversion or specified units in existed header file (based on your opinion).
- [x] parse `self.time` from` vtk.out`
- [x] add new tests


neutrinoceros:
Thank you for your work !
I'll review thoroughly later, but for now here's an important thing that I didn't document explicitly:
This repo uses pre-commit to check/format code style. It's following yt's guideline, see https://yt-project.org/docs/dev/developing/developing.html#coding-style-guide

xshaokun:
I just installed pre-commit and ran it. However, I need your help for some error reports.

``` bash
yt_idefix/fields.py:7: error: Need type annotation for "known_other_fields"
yt_idefix/fields.py:25: error: Need type annotation for "known_other_fields"
```

These raised from your code. Should I just ignore it?

```bash
yt_idefix/data_structures.py:19:1: F401 '.definitions.PlutoDefConstants' imported but unused
```

`PlutoDefConstants` is a class I defined to store some contants, and set all constants as its attributes. Then I replace the constant in the expression of code_unit (refer to the new `tests/data/pluto_sedov/definitions.h`) through `expr.replace("CONST", "PlutoDefConstants().CONST")`. The expression will be `eval` later. That's why the error raised. The `PlutoDefConstants` is used as a string.

Is there any way to ignore it? Or do you know a better way to do it.

xshaokun:
I resolved the second F401 error through commenting ` #noqa` at the end of that line and pre-commit ignored it.

neutrinoceros:
> These raised from your code. Should I just ignore it?

Yes, the correct type here is a little hard to get, but it is defined on yt's main branch, which is why I haven't added annotation yet. You can deactivate mypy locally with `# type: ignore`, but please do not abuse it ! it's 100% fine for the two lines at play here.

> I resolved the second F401 error through commenting #noqa at the end of that line and pre-commit ignored it.

Ok that will do for now. I think we could improve the design here but for now let's focus on getting your workflow set up :)

xshaokun:
> You can deactivate mypy locally with `# type: ignore`, but please do not abuse it ! it's 100% fine for the two lines at play here.

such comments have been already added here, but somhow they didn't work.

xshaokun:
> such comments have been already added here, but somhow they didn't work.

The comments were added for wrong lines.

neutrinoceros:
Pre-commit.ci is now happy, well done !
Now the next thing we need to do is making sure that you don't break existing tests. Right now there are some failures, so please fix them before anything else.
Tests can be run locally with `pytest`, and I suggest you make sure that everything stays green on your local machine before you push again, because currently I need to approve every time you push (this is because it's your first patch to the repo).

I'll make a thorough review after that !
Again, thank you for spending the effort, it's really appreciated !

neutrinoceros (commented):
here are a couple quick comments.

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,77 @@ def _parse_parameter_file(self):
...
+                for line in hf.readlines():
+                    if 'GEOMETRY' in line:
+                        self.geometry = line.split()[-1].lower()
+                    if 'UNIT' in line:
+                        item = line.split()
+                        key = item[1].lower()
+                        expr = item[-1]
+                        if 'CONST' in expr:
+                            expr = expr.replace('CONST', 'PlutoDefConstants().CONST')
+                        self._specified_code_units[key] = eval(expr)
```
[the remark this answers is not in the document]
xshaokun: As you see, the code_unit could be defined by an expression in Pluto. The only way I know to parse the expression is `eval`. I can specify the unit with a number in my own Pluto code, so there’s no need for `eval`. But I presumed this package would be used by other Pluto users. Do you have any better ideas?

xshaokun:
Some errors reported `pytest` :

```
_____________________________________________________________________ ERROR collecting tests/test_vtk.py ______________________________________________________________________
tests/test_vtk.py:8: in <module>
    from yt.extensions.idefix.api import IdefixVtkDataset, PlutoVtkDataset
yt_idefix/api.py:2: in <module>
    from .data_structures import (
yt_idefix/data_structures.py:106: in <module>
    class IdefixDataset(Dataset, ABC):
../anaconda3/lib/python3.7/abc.py:126: in __new__
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
../yt/yt/data_objects/static_output.py:190: in __init_subclass__
    f"Overwritting {cls.__name__}, which was previously registered. "
E   UserWarning: Overwritting IdefixDataset, which was previously registered. This is expected if you're importing a yt extension with a frontend that was already migrated to the main code base.
=========================================================================== short test summary info ===========================================================================
ERROR tests/test_dmp.py - UserWarning: Overwritting IdefixDataset, which was previously registered. This is expected if you're importing a yt extension with a frontend that...
ERROR tests/test_vtk.py - UserWarning: Overwritting IdefixDataset, which was previously registered. This is expected if you're importing a yt extension with a frontend that...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 2 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

Above error reports raised with `yt-4.1.dev0`(github repo version). Then I switched to stable version provided by `conda`:

```
tests/test_vtk.py::test_load_magic[fn0-None] XFAIL (all .vtk files are detected as Athena-compatible, hence dataformat is considered ambiguous before yt 4.0.2)
tests/test_vtk.py::test_load_magic[fn1-polar] XFAIL (all .vtk files are detected as Athena-compatible, hence dataformat is considered ambiguous before yt 4.0.2)
tests/test_vtk.py::test_load_magic[fn2-spherical] XFAIL (all .vtk files are detected as Athena-compatible, hence dataformat is considered ambiguous before yt 4.0.2)
tests/test_vtk.py::test_load_magic[fn3-cartesian] XFAIL (all .vtk files are detected as Athena-compatible, hence dataformat is considered ambiguous before yt 4.0.2)
...
======================================================================= 25 passed, 4 xfailed in 36.64s ========================================================================
```

Should I ignore the xfails?

tests/data/pluto_sedov/definitions.h
```diff
@@ -27,5 +27,8 @@
 /* [Beg] user-defined constants (do not change this line) */
 
 #define  INITIAL_SMOOTHING              YES
+#define  UNIT_LENGTH                    1000*CONST_pc
+#define  UNIT_VELOCITY                  1e5
+#define  UNIT_DENSITY                   1000*CONST_mp
```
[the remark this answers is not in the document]
xshaokun: Sure! it's better to make it clear.

yt_idefix/definitions.py
```diff
@@ -0,0 +1,35 @@
+class PlutoDefConstants:
+    """Physical constants in c.g.s units defined in $PLUTO_DIR/Src/pluto.h"""
+
+    __pluto_def_constants = {
```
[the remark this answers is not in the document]
xshaokun: The class is defined for conveniently replace the constants in the code unit expression, then `eval`. Apparently I need to redesign this unit part later. For now I will drop the socond underscore in the next commit.

neutrinoceros:
Thanks for reporting this. Indeed there are a couple of incompatibilities at the moment with the dev branch of yt that I wasn't aware of (even though it seems that I caused them myself). I will fix these as soon as possible but for now, I advise you use a stable release in your yt_idefix dev environment.

> Should I ignore the xfails?

yes. By definition `xfail` means "expected to fail", these are tests that are suppose to work in future versions of yt but don't with the current latest.

xshaokun:
Except for the `eval`, I've done what I can do so far.

### PR Summary

- Parse geometry and code units from header file `definitions.h`, which is searched from directory of data file or  code directory specified.
- Parse current time from `vtk.out` which should be along with data file.

Welcome your review and comments :)

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,98 @@ def _parse_parameter_file(self):
...
     _required_header_keyword = "PLUTO"
+
+    def __init__(
+        self,
+        filename,
+        dataset_type=None,
+        unit_system="cgs",
+        units_override=None,
+        *,
+        code_dir: str | None = None,
```
neutrinoceros, line 332: Unless you have ideas how this argument could be used for other things in the future, I'd much prefer exposing a `definitions_header` argument directly, which would match how the `inifile` argument is implemented. The corresponding instance attribute should be "hidden" rather than public (`self._definitions_header` instead of `self.code_dir`)

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,98 @@ def _parse_parameter_file(self):
...
+                for line in hf.readlines():
+                    if "GEOMETRY" in line:
+                        self.geometry = line.split()[-1].lower()
+                    if "UNIT" in line:
+                        item = line.split()
+                        key = item[1].lower()
+                        expr = item[-1]
+                        if "CONST" in expr:
+                            expr = constexpr.sub(self._get_const_for_reg, expr)
+                        self.parameters[key] = eval(expr)
```
neutrinoceros, line 372: the implementation is rather confusing to me. I don't think I can maintain that without rewriting/simplifying it. I'm fine with refactoring it myself later but I'll let you the initiative if you feel like doing it.

In particular, it's a little unsettling that pluto_def_constants' values are floats, but you're only ever using them as `str`, this feels like it could be improved.

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,98 @@ def _parse_parameter_file(self):
...
+            )
+
+    def _get_const_for_reg(self, matched):
+        """Replace matched constant string with value"""
+        key = matched.group()
+        return str(pluto_def_constants[key])
+
+    def _get_time(self):
+        """Get current time from vtk.out."""
+        log_file = os.path.join(self.directory, "vtk.out")
```
neutrinoceros, line 392: Is `"vtk.out"` a universal name for this file ? If not, it should probably be exposed

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,98 @@ def _parse_parameter_file(self):
...
+        index = int(re.search(r"\.(\d*)\.", self.parameter_filename).group(1))
```
neutrinoceros, line 393: I'm surprised mypy didn't complain about that
```suggestion
        match = re.search(r"\.(\d*)\.", self.parameter_filename)
        if match is None:
            raise RuntimeError(
                f"Failed to parse output number from file name {self.parameter_filename}"
            )
        index = int(match.group(1))
```

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,98 @@ def _parse_parameter_file(self):
...
+        if os.path.isfile(log_file):
+            with open(log_file) as log:
+                for line in log.readlines():
+                    if line.startswith(str(index)):
+                        self.current_time = np.float32(line.split()[1])
+        else:
+            warnings.warn(
+                f"Log file couldn't be found in {self.fullpath}. "
+                "The 'current_time' attributte is set to be -1."
+            )
```
neutrinoceros, line 403: this doesn't mandate as a `warnings.warn` call in my opinion because it's unlikely that users getting the warning will be able to act on it.
I'd go with silently setting the default value (which seems to be missing here actually)

yt_idefix/fields.py
```diff
@@ -3,8 +3,8 @@
 
 
 class IdefixVtkFieldInfo(FieldInfoContainer):
-    known_other_fields = (  # type: ignore
-        ("RHO", ("code_mass / code_length**3", ["density"], None)),
+    known_other_fields = (
+        ("RHO", ("code_mass / code_length**3", ["density"], None)),  # type: ignore
```
neutrinoceros, line 7: Probably equivalent, slightly more elegant in my opinion
```suggestion
    known_other_fields = (   # type: ignore
        ("RHO", ("code_mass / code_length**3", ["density"], None)),
```

yt_idefix/fields.py
```diff
@@ -21,8 +21,8 @@ def setup_fluid_fields(self):
 
 
 class IdefixDmpFieldInfo(FieldInfoContainer):
-    known_other_fields = (  # type: ignore
-        ("Vc-RHO", ("code_mass / code_length**3", ["density"], None)),
+    known_other_fields = (
+        ("Vc-RHO", ("code_mass / code_length**3", ["density"], None)),  # type: ignore
```
neutrinoceros, line 25: same as above
```suggestion
    known_other_fields = (   # type: ignore
        ("Vc-RHO", ("code_mass / code_length**3", ["density"], None)),
```

neutrinoceros (changes requested):
Congrats on getting CI green !
I think your contribution is of excellent quality and I will definitely end up merging it soon, but I'd like to improve it furthermore so it's even more stable.

I'm requesting only one API change here, and the rest of my comments are internal stuff.
We'll need tests for your contribution too. Do you feel comfortable writing some ?
If not, I'll take over your PR once you're done with it and add tests myself.

Thank you for your patience and enthusiasm, I'm really looking forward to publishing the next version with your changes incorporated !

yt_idefix/data_structures.py
xshaokun: make sense! I'll change it

yt_idefix/data_structures.py
xshaokun: I should have left more comments in this part. For the cases where the pre-defined constant is used to define code units, such as `1000*CONST_pc`, the expression is a string and I replace the constant with its value. Then the code unit is able to be calculated. 

This might be a stupid way but works. I'm wiiling to rewrite it if there's a better way.

> In particular, it's a little unsettling that pluto_def_constants' values are floats, but you're only ever using them as `str`, this feels like it could be improved.

The values stored as floats is intuitive although I use them as `str` later. I can stored them as `str` if necessary.

yt_idefix/data_structures.py
xshaokun: Yes. In Pluto documents:

> Each format has an independent output frequency and an associated log ﬁle (i.e. dbl.out, ﬂt.out, vtk.out and so forth) keeping track of the dump history.

And there's no API to change it unless modify source code.

yt_idefix/data_structures.py
xshaokun: Thanks, it's more readable!

yt_idefix/data_structures.py
xshaokun: > it's unlikely that users getting the warning will be able to act on it.

Although the value of `current_time` is printed by yt_logger after `load`, I thought it can tell users the `current_time` isn't set correctly due to lacking log file, not due to other bugs.

> I'd go with silently setting the default value (which seems to be missing here actually)

Because `_get_time` is ran after `super()._parse_parameter_file()`, the default value `-1` has already been set.

yt_idefix/fields.py
xshaokun: Unfortunately, your original location of ` # type: ignore` couldn't pass `mypy` in my machine.

xshaokun:
Thanks for your comments here and before ! I've learned a lot about coding style! 

> We'll need tests for your contribution too. Do you feel comfortable writing some ?

I don't know how to write tests. If it's not urgent I'd like to learn it and write them by myself. Please tell me what I need to do if you think it's okay :)

yt_idefix/fields.py
neutrinoceros: ok fair enough, thanks for checking !

neutrinoceros:
> I don't know how to write tests. If it's not urgent I'd like to learn it and write them by myself. Please tell me what I need to do if you think it's okay :)

no urgency on my side, feel free to work through this at your own pace.
I recommend reading an introduction to pytest. The following links may be good places to start, though I have not read them myself
https://docs.pytest.org/en/6.2.x/getting-started.html
https://realpython.com/pytest-python-testing/

I also recommend the excellent pytest book, which is the de facto bible on the topic. The second edition is currently in beta so it should be accessible for free, at least in part here https://pragprog.com/titles/bopytest2/python-testing-with-pytest-second-edition/

Otherwise, taking inspiration from existing tests (in `tests/`) is also a good idea but I think it's pretty hard to understand how they work just by reading examples, which is why I recommend reading some documentation first.

yt_idefix/data_structures.py
neutrinoceros: Excellent, thank you ! Nothing to do here then

yt_idefix/data_structures.py
neutrinoceros: the main problem I see with your implementation is that it may turn into unspecified behaviours with valid data

For instance
```python
for line in hf.readlines():
    if "GEOMETRY" in line:
        self.geometry = line.split()[-1].lower() 
```

would produce problematic results with the following content
```C
#define GEOMETRY CARTESIAN
// GEOMETRY could be set to POLAR
```
with your current implementation, we'd get `self.geometry = "polar"` while the correct result should be `"cartesian"`

We should parse it in a much more restricted fashion to avoid this bug, for instance
```python
geom_regexp = re.compile(r"#define\s+GEOMETRY\s+([A-Z]+)")
for line in lines:
    match = re.search(geom_regexp, line)
    if match is None:
        continue
    self.geometry = match.group(1).lower()
```

ignoring multi-lines C comments would also be useful but it's harder.

xshaokun: I see. I'll rewrite it, thanks!

yt_idefix/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: This thread is outdated so I'll close it, but yes, I'm convinced we can avoid using `eval`, and I'll help you design a solution soon if needed

yt_idefix/data_structures.py
neutrinoceros: btw, I see you used `hf` instead of `fh` (file handle) as variable name

yt_idefix/data_structures.py
neutrinoceros: > Although the value of current_time is printed by yt_logger after load, I thought it can tell users the current_time isn't set correctly due to lacking log file, not due to other bugs.

if you want to warn on stuff that users can't act upon, please use the yt logger itself. `warnings.warn` should be reserved to problems that users are able to fix on their end. Not having a log file doesn't seem like something users could fix.

neutrinoceros: > Because _get_time is ran after super()._parse_parameter_file(), the default value -1 has already been set.

True, but it's not obvious from looking at the code here. It could also be a different value if something changes upstream, in which case your warning message would be wrong. You could mitigate this by inserting the value at runtime.

yt_idefix/data_structures.py
xshaokun: `hf` is short for header file. Is it for convension to use `fh`?

yt_idefix/data_structures.py
xshaokun: Okay, now i get it.

yt_idefix/data_structures.py
neutrinoceros: In yt (and by extension yt_idefix), the dominant convention is `fh` for file handle whatever the file, I'd prefer to keep it consistent.

neutrinoceros: I thought about the problems with the accidentally parsing comments and my solution isn't actually much more robust than yours. The difficulty comes from that we're parsing the header file line by line, whereas it would be much easier to do it correctly if we stripped out comments first.

xshaokun: Inspiring by your example of geometry, now I use regular expression to identify the true expression. I think the comments won't be accidentally included.

neutrinoceros: How about this ?
```
#define GEOMETRY CARTESIAN
//#define GEOMETRY POLAR
```
And it gets worse with multi-line comments.
Anyway, that's what tests are for :)

xshaokun: I use `"^\s*#define\s+GEOMETRY\s+([A-Z]+)"` in the new commit. It can ignore any comments, I think.

neutrinoceros: That'll do

tests/test_vtk.py
```diff
@@ -35,6 +36,22 @@ def test_validation_pluto(fn):
...
 
+@pytest.mark.parametrize(("fn", "geometry"), PLUTO_VTK_FILES)
+def test_parse_pluto_metadata(fn, geometry):
+    ds = yt_idefix.load(fn)
+    assert ds._definitions_header == os.path.join(ds.fullpath, "definitions.h")
+    assert "unit_length" in ds.parameters
+    assert "unit_density" in ds.parameters
+    assert "unit_velocity" in ds.parameters
+    assert ds.current_time != -1
+    assert ds.geometry == "cartesian"
```
neutrinoceros, line 47:
```suggestion
    assert ds.geometry == geometry
```

tests/test_vtk.py
```diff
@@ -35,6 +36,22 @@ def test_validation_pluto(fn):
...
 
 
+@pytest.mark.parametrize(("fn", "geometry"), PLUTO_VTK_FILES)
+def test_parse_pluto_metadata(fn, geometry):
+    ds = yt_idefix.load(fn)
+    assert ds._definitions_header == os.path.join(ds.fullpath, "definitions.h")
+    assert "unit_length" in ds.parameters
+    assert "unit_density" in ds.parameters
+    assert "unit_velocity" in ds.parameters
+    assert ds.current_time != -1
```
neutrinoceros, line 46:
```suggestion
    assert ds.current_time >= 0
```

tests/test_vtk.py
```diff
@@ -35,6 +36,22 @@ def test_validation_pluto(fn):
...
+    assert ds.geometry == "cartesian"
+    assert ds.length_unit / ds.velocity_unit == ds.time_unit
+    with pytest.raises(ValueError):
+        ds = yt_idefix.load(fn, definitions_header="definitions2.h")
+    with pytest.warns(UserWarning):
+        ds = yt_idefix.load(fn, definitions_header="definitions2.h", geometry=geometry)
```
neutrinoceros, line 52: a couple remarks here
1) these looks like 2 different tests, it'd be preferable to separate them from the main one, for a total of 3 test functions
2) please use the `match` argument in `pytest.warns` and `pytest.raises` to match the string you're expecting. If it's an exact string, it should be straighforward, but beware that `match=` expects a regexp pattern in general. Here again, I'm happy to do it for you if you can at least tell me what the expected errors and warnings look like.

neutrinoceros (commented):
Great job on writing this test ! It's impressive how little I have to say about it, for a first time writing a test this is excellent work.

I also see that you're not checking which specific units end up being attached to the dataset. 
Could you also add a test to do that with the pluto_sedov dataset, since you've added values there ?

We're getting really close to a merge-ready state, I'll do a (hopefully last) round of review once you are done with tests :)

tests/test_vtk.py
xshaokun: Here I want to test the occasions when some arguments are missed or assigned incorrectly, which are expected in my `if-elif-else` statements in `_parse_header_file()`. The exact strings for `ValueError` and `warnings` are written in `_parse_header_file()`. So I think maybe it's verbose to write them here again?

xshaokun: I just got your point! I'll improve it to make the test more accurate. Thanks!

> 1. these looks like 2 different tests, it'd be preferable to separate them from the main one, for a total of 3 test functions

Besides, as I said, these two tests are designed for two expected cases during parsing metadata, should I separate them?

neutrinoceros: > So I think maybe it's verbose to write them here again?

It doesn't matter if it's verbose, the role such a test is to ensure we're getting the expected error. Matching the string ensures we're not passing the test by accident because some _different_ `ValueError` is raised. Besides, tests read like a form of in depth documentation, if I can't tell what's the error without looking at the current implementation then it's more of a pain to maintain.
In general, test code doesn't have to be pretty, or minimal, or [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself); what's important is that it's clear and as maintainable as possible

xshaokun: I see. Thanks for your explainations again! I've learned a lot from this work. I wrote these test just after knowing some basic usage and I'll learn more details later!

neutrinoceros: > Besides, as I said, these two tests are designed for two expected cases during parsing metadata, should I separate them?

Yes, it's okay to have several tests related in topics. What's important here is that we avoid shadowing a failure with another in case both are broken at the same time

tests/test_vtk.py
```diff
@@ -39,17 +40,45 @@ def test_validation_pluto(fn):
...
     assert "unit_velocity" in ds.parameters
-    assert ds.current_time != -1
-    assert ds.geometry == "cartesian"
-    assert ds.length_unit / ds.velocity_unit == ds.time_unit
-    with pytest.raises(ValueError):
-        ds = yt_idefix.load(fn, definitions_header="definitions2.h")
-    with pytest.warns(UserWarning):
-        ds = yt_idefix.load(fn, definitions_header="definitions2.h", geometry=geometry)
+    assert ds.geometry == geometry
+    assert (
```
neutrinoceros, line 50: Calls to `assert_allclose_units` don't need to be chained with plain `assert`. Please simplify.

Another problem here is that this is a parametrised test. Right now we have only one Pluto dataset, so it works, but what you're testing here is specific to this dataset, so it should be done in a separate test

tests/test_vtk.py
```diff
@@ -39,17 +40,45 @@ def test_validation_pluto(fn):
...
+        assert_allclose_units(ds.time_unit, unyt_quantity(3.08567758e16, "s")) is None
+    )
+    assert (
+        assert_allclose_units(ds.magnetic_unit, unyt_quantity(4.58462477e-07, "G"))
+        is None
+    )
+    assert ds.current_time >= 0
+
+
+@pytest.mark.parametrize("fn", [_[0] for _ in PLUTO_VTK_FILES])
```
neutrinoceros, line 70: No need to parametrize this test, I would rather hardcode the name of one dataset here.

tests/test_vtk.py
```diff
@@ -39,17 +40,45 @@ def test_validation_pluto(fn):
...
+def test_pluto_value_error(fn):
+    error_msg = "Header file definitions2.h couldn't be found. The 'geometry' keyword argument must be specified."
+    with pytest.raises(ValueError, match=error_msg):
+        yt_idefix.load(fn, definitions_header="definitions2.h")
+
+
+@pytest.mark.parametrize(("fn", "geometry"), PLUTO_VTK_FILES)
```
neutrinoceros, line 77: same as above

tests/test_vtk.py
xshaokun: Okay. How about creating `PLUTO_SEDOV` at header for those tests where checking one dataset is enough, also in case future changes on `pluto_sedov`

neutrinoceros: Sure, go ahead

xshaokun:
Maybe I should ask this question before this commit. `assert_allclose_units` alone will trigger flake8/B015. I just ignored it by` #noqa: B015` at those lines. Do you know a better way to do it? I found yt just call it in its tests.

neutrinoceros:
> Maybe I should ask this question before this commit. assert_allclose_units alone will trigger flake8/B015. I just ignored it by #noqa: B015 at those lines. Do you know a better way to do it? I found yt just call it in its tests.

It's very surprising. I checked in your branch, removed these noqa comments, ran `pre-commit run --all-files` and got no such error. I also don't recall ever having to disable flake8-bugbear in similar tests in yt.
Frankly I don't understand why this breaks on your machine. A simple (yet annoying) workaround you could either uninstall `pre-commit` (pre-commit.ci will still lint your code and possibly auto fix it when you push), or use the `--no-verify` flag in `git commit` to skip pre-commit.

tests/test_vtk.py
```diff
@@ -22,6 +24,8 @@
 )
 PLUTO_VTK_FILES = ((DATA_DIR.joinpath("pluto_sedov", "data.0018.vtk"), "cartesian"),)
 
+PLUTO_SEDOV = DATA_DIR.joinpath("pluto_sedov", "data.0018.vtk")
```
neutrinoceros, line 27:
```suggestion
PLUTO_SEDOV = DATA_DIR.joinpath("pluto_sedov", "data.0018.vtk")

PLUTO_VTK_FILES = ((PLUTO_SEDOV, "cartesian"),)
```

yt_idefix/definitions.py
```diff
@@ -0,0 +1,29 @@
+# Physical constants in c.g.s units defined in $PLUTO_DIR/Src/pluto.h
+pluto_def_constants = {
```
neutrinoceros, line 2: I'll probably need to update this in the future if upstream (Pluto) changes a value or add new ones. Could you please document (in a comment) which version of Pluto you used as a reference here ?

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,108 @@ def _parse_parameter_file(self):
...
+            raise RuntimeError(
+                f"Failed to parse output number from file name {self.parameter_filename}"
+            )
+        index = int(match.group(1))
+
+        if os.path.isfile(log_file):
+            with open(log_file) as fh:
+                for line in fh.readlines():
+                    if line.startswith(str(index)):
+                        self.current_time = np.float32(line.split()[1])
```
neutrinoceros, line 407: A small improvement here: no need to keep reading after we find the value.
Also, I suggest raising an error (or possibly warning ?) in case this doesn't work. I leave this up to up and left a `...` where the error should be.
Note that `else` blocks in for loops read as "if no break".
```suggestion
                        self.current_time = np.float32(line.split()[1])
                        break
                else:
                    ...
```

yt_idefix/definitions.py
xshaokun: Sure

xshaokun:
> removed these noqa comments, ran `pre-commit run --all-files` and got no such error.

It's really wierd. I followed the same steps this time and no error appeared. Maybe I changed something unwarily but it's multiple lines. Now it's fine. I'll remove those comments in the next commit.

yt_idefix/data_structures.py
xshaokun: Make sense. It's better.

xshaokun: I just found a bug when parse the value. So I'll use regular expression in the next commit to make it more secure. I've tested it locally, though I think it would be better to have a test for this part in the future (no good idea now).

neutrinoceros: Interesting. I didn't suggest using a regexp because I didn't have any idea how to break your implementation. Indeed if you know how to break it, it's worth adding a test. I might be able to make some suggestions how to if you can describe the bug first

xshaokun: It's a very wierd case ideed. When some information is removed manually in the log file, for example, the info of first snapshot is missing, `startswith(1)` would locate 10th snapshot info. This case must be ridiculous.

Actually the test I mentioned is for those `else-then-warning` part.

xshaokun: Do you think use a regexp here is too expansive? like this:

``` python
 for line in fh.readlines():
     log_match = re.search(log_regexp, line)
     if log_match.group(1) == str(index):
         self.current_time = np.float64(log_match.group(2))
```

neutrinoceros: Oh I see. Well, manually editing a log file sounds like a terrible idea, _but_ you never know, and it's conceivable that a log file would end up corrupted e.g. because of a bug in Pluto.
You could write a very simple test case where the log file would exist but be empty.

neutrinoceros: Damn we overlapped. Yes a regexp is fine but you must also check that you actually get a match and not `None` before you retrieve a group.

xshaokun: Just see it. Please ignore this commit. I'll amend it.

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,116 @@ def _parse_parameter_file(self):
...
+        index = int(match.group(1))
+
+        self.current_time = -1
+        if os.path.isfile(log_file):
+            log_regexp = re.compile(fr"^{index}\s(\S+)")
+            with open(log_file) as fh:
+                for line in fh.readlines():
+                    log_match = re.search(log_regexp, line)
+                    if log_match:
+                        self.current_time = np.float64(log_match.group(1))
```
neutrinoceros, line 410: Note that `self.current_time` should be a unyt_quantity object, not a float. I think this is how you need to write it
```suggestion
                        self.current_time = self.quan(float(log_match.group(2)), "code_time")
```

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,116 @@ def _parse_parameter_file(self):
...
+        match = re.search(r"\.(\d*)\.", self.parameter_filename)
+        if match is None:
+            raise RuntimeError(
+                f"Failed to parse output number from file name {self.parameter_filename}"
+            )
+        index = int(match.group(1))
+
+        self.current_time = -1
+        if os.path.isfile(log_file):
+            log_regexp = re.compile(fr"^{index}\s(\S+)")
```
neutrinoceros, line 405: if I'm not mistaken, the second capturing group `(\S+)` is meant to represent a number (time) possibly written in scientific notation, right ?
Eventually I'd like to refactor this to utilise `inifix.ENotationIO.decode`, but I'm not requiring this for this PR (inifix is a different library that I also maintain and which is already a direct dependency to yt_idefix).

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,116 @@ def _parse_parameter_file(self):
...
+            with open(log_file) as fh:
+                for line in fh.readlines():
+                    log_match = re.search(log_regexp, line)
+                    if log_match:
+                        self.current_time = np.float64(log_match.group(1))
+                        break
+                if self.current_time == -1:
```
neutrinoceros, line 412: Logically it should be equivalent, unless if for whatever reason "-1" becomes a valid time value for whatever reason in the future, in which case your current implementation is broken
```suggestion
                else:
```

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,116 @@ def _parse_parameter_file(self):
...
+                    mylog.warning(
+                        "The snapshot information cannot found in log file vtk.out. "
+                        "The 'current_time' attributte is set to be -1."
+                    )
```
neutrinoceros, line 416:
```suggestion
                    mylog.warning(
                        f"Failed to retrieve time from {log_file}, setting current_time = -1"
                    )
```

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,116 @@ def _parse_parameter_file(self):
...
+        else:
+            mylog.warning(
+                "Log file vtk.out couldn't be found. "
+                "The 'current_time' attributte is set to be -1."
+            )
```
neutrinoceros, line 421:
```suggestion
            mylog.warning(
                f"Missing log file {log_file}, setting current_time = -1"
            )
```

tests/test_vtk.py
```diff
@@ -35,6 +42,40 @@ def test_validation_pluto(fn):
...
+    assert_allclose_units(ds.velocity_unit, unyt_quantity(1e5, "cm/s"))
+    assert_allclose_units(ds.mass_unit, unyt_quantity(4.91416082e40, "g"))
+    assert_allclose_units(ds.time_unit, unyt_quantity(3.08567758e16, "s"))
+    assert_allclose_units(ds.magnetic_unit, unyt_quantity(4.58462477e-07, "G"))
+
+
+def test_pluto_value_error():
+    error_msg = "Header file definitions2.h couldn't be found. The 'geometry' keyword argument must be specified."
+    with pytest.raises(ValueError, match=error_msg):
+        yt_idefix.load(PLUTO_SEDOV, definitions_header="definitions2.h")
```
neutrinoceros, line 68: Actually now that you added the actual message I think the error should be a `FileNotFoundError` rather than `ValueError`

neutrinoceros (changes requested):
I think this will be my last round of comments. I'll merge when you're done dealing with these.
Feel free to squash your commit into a single one if you want to (only suggesting this because you seem to be comfortable with rebasing, this isn't a required step!)

yt_idefix/data_structures.py
xshaokun: > if I'm not mistaken, the second capturing group (\S+) is meant to represent a number (time) possibly written in scientific notation, right ?

Yes, you're right. I checked inifix and it seems like `inifix.ENotationIO.decode` cannot be called outside the package right now. So I'll leave it for future PR.

yt_idefix/data_structures.py
xshaokun: Exactly! Btw, the regexpr here only captured one string (only one pair of parentheses), so it should be `group(1)` and `group(2)` doesn't exist.

yt_idefix/data_structures.py
xshaokun: Oh I just realized that `for` statement can be followed by `else`, knowing a new feature of python. I misunderstand your explaination before, my bad.

tests/test_vtk.py
xshaokun: make sense!

xshaokun: > yt_idefix/data_structures.py:416:25: G004 Logging statement uses f-string

However, f-string is not suggeted in logging statements by flake8. Should I ignore it, or do it like this?

``` python
warn_msg = f"Failed to retrieve time from {log_file}, setting current_time = -1"
mylog.warning(warn_msg)
```

yt_idefix/data_structures.py
xshaokun: > yt_idefix/data_structures.py:416:25: G004 Logging statement uses f-string

However, f-string is not suggeted in logging statements by flake8. Should I ignore it, or do it like this?

``` python
warn_msg = f"Failed to retrieve time from {log_file}, setting current_time = -1"
mylog.warning(warn_msg)
```

yt_idefix/data_structures.py
neutrinoceros: No worries, I'll do it myself. Possibly I'll change how inifix exposes it.

yt_idefix/data_structures.py
neutrinoceros: It's a little obscure, and underused, but yeah it's valid :)

yt_idefix/data_structures.py
neutrinoceros: Oh no that's right, don't ignore it. I forgot that formatted logging statements require a different syntax, for instance
```python
mylog.error("invalid value: %s", val)
```
This is because evaluating an fstrings can be costly, and we don't want to waste resources evaluating strings that won't be printed, so logging statements have this special syntax to postpone their evaluation until we actually need it, depending on the runtime logging level

xshaokun: Got it!

xshaokun:
Hope I finish it this time. I didn't sqush my commits, thinking it would be easier for you to review thoroughly. Squashing them when merging would be better, perhaps.

Anyway, I sincerely appreciate your help and encouragement during this time. You have been patient to guide me, surly doing this work by yourself would be more effecient though. I've learned a lot indeed, especially about how to maintaining a package or module. 

Hope I could make more contributions in the future, with higher-quality coding of course :)

yt_idefix/data_structures.py
```diff
@@ -320,3 +321,117 @@ def _parse_parameter_file(self):
...
+                        )
+                        break
+                else:
+                    mylog.warning(
+                        "Failed to retrieve time from %s, setting current_time = -1",
+                        log_file,
+                    )
+        else:
+            mylog.warning(
+                "Missing log file {log_file}, setting current_time = -1", log_file
```
neutrinoceros, line 421:
```suggestion
                "Missing log file %s, setting current_time = -1", log_file
```

xshaokun: Sorry, so reckless!

neutrinoceros:
> Hope I finish it this time. I didn't sqush my commits, thinking it would be easier for you to review thoroughly. Squashing them when merging would be better, perhaps.

Yes it's easier on the reviewer to avoid squashing repeatedly during the review process, however since we're essentially done now, it's okay if you want to do it. Otherwise I'll squash on merge.

> Anyway, I sincerely appreciate your help and encouragement during this time. You have been patient to guide me, surly doing this work by yourself would be more effecient though. I've learned a lot indeed, especially about how to maintaining a package or module.

I'm glad you enjoyed the experience, and I'm looking forward to seeing more contributions from you either here on in yt itself if you ever need to. Again, I'm impressed at how much you've learned from a single PR.  Getting competent contributors is extremely valuable so I consider my time was well invested here.

 I'll be back in an hour or so, I'll merge regardless the state (squashed or not) by then, and make a release today.

xshaokun:
- [ ] Got squash skill :)

neutrinoceros (approved):

PR #68, opened 2022-01-16
BUG: fix chunk index problem

### LINKED ISSUE #67: BUG: The value of derived fields cannot update correctly
Here is a simple piece of code for reproducing the bug quickly.

```Python
import yt
import yt_idefix

from yt.utilities.physical_constants import mh, kb

sedov = "pluto_sedov/data.0018.vtk"
ds = yt_idefix.load(sedov, unit_system="cgs")

def temp(field, data):
    mu = 0.6
    return mu * mh * data[("gas","pressure")] / data[("gas","density")] / kb
ds.add_field(name=("gas","temperature"),function=temp, units="K",sampling_type="local")

ad = ds.all_data()
print("added_temperature = ", ad["gas","temperature"])
```

The result showed that the value of temperature is infinity or other wrong number, due to the wrong value of [density]. Using `("idefix-vtk","RHO)` and `("idefix-vtk","PRS)` doesn't make differences. I also tried different ways to add the derived_field, e.g. `@yt.derived_field`

```bash
yt : [INFO     ] 2022-01-15 16:50:59,762 Parameters: current_time              = 0.02279213 code_time
yt : [INFO     ] 2022-01-15 16:50:59,763 Parameters: domain_dimensions         = [48 48 48]
yt : [INFO     ] 2022-01-15 16:50:59,763 Parameters: domain_left_edge          = [0. 0. 0.]
yt : [INFO     ] 2022-01-15 16:50:59,764 Parameters: domain_right_edge         = [1. 1. 1.]
yt : [INFO     ] 2022-01-15 16:50:59,764 Parameters: cosmological_simulation   = 0
/share/shaokun/Apps/anaconda3/lib/python3.7/site-packages/unyt/array.py:1779: RuntimeWarning: divide by zero encountered in true_divide
  inp0.view(np.ndarray), inp1.view(np.ndarray), out=out_func, **kwargs
added_temperature =  [-2.54617343e-45  4.49557643e-49 -1.48152055e-97 ...             inf
             inf             inf] K
```

I found that calling [density] field before [temperature] would make [temperature] right, such as `print(ad["gas","density"].

I made the same test for FLASH data (by force-overriding the original temperature function) and it works. So I think the bug might come from this extension instead of yt, or might not.

In my understanding, the information of derived fields are stored in `FieldInfoContainer` after adding them, with their dependent field arrays whose value were generated randomly and only update once they are called. So the bug seems like that the initial array was used when calculating the derived field. I tried to figure out when the array was updated but failed.

xshaokun:
### PR Summary
- fixes #67 
- adds test for derived field

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = yt_idefix
-version = 0.9.2
+version = 0.9.3
 description = An extension module for yt, adding a frontend for Idefix
 long_description = file: README.md
 long_description_content_type = text/markdown
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -102,6 +102,27 @@ def test_load_parseable_geometry_metadata(vtk_file_with_geom):
     assert ds.geometry == file["geometry"]
 
 
+def test_derived_field(vtk_file):
+    file = vtk_file
+    ds = yt_idefix.load(file["path"], geometry=file["geometry"])
+    # this derived field is compatible for all current test data
+    # it'll be replaced once a better universal field is defined internally.
+
+    def mom_den(field, data):
+        return data["idefix-vtk", "RHO"] * data["idefix-vtk", "VX1"]
+
+    ds.add_field(
+        name=("gas", "momentum_density"),
+        function=mom_den,
+        units="g*cm**-2/s",
+        sampling_type="cell",
+    )
+    ad = ds.all_data()
+    test = ad["gas", "momentum_density"]
+    expect = ad["idefix-vtk", "RHO"] * ad["idefix-vtk", "VX1"]
+    assert_allclose_units(test, expect)
+
+
 # TODO: make this a pytest-mpl test
 def test_slice_plot(vtk_file):
     file = vtk_file
--- a/yt_idefix/__init__.py
+++ b/yt_idefix/__init__.py
@@ -2,4 +2,4 @@
 # immediately after `import yt.extensions.idefix`
 from yt_idefix.api import *
 
-__version__ = "0.9.2"
+__version__ = "0.9.3"
--- a/yt_idefix/io.py
+++ b/yt_idefix/io.py
@@ -19,17 +19,21 @@ class IdefixIOHandler(BaseIOHandler, ABC):
 
     def _read_fluid_selection(self, chunks, selector, fields, size):
         data = {}
-        offset = 0
+
+        for field in fields:
+            data[field] = np.empty(size, dtype="float64")
 
         with open(self._data_file, mode="rb") as fh:
-            for field in fields:
-                ftype, fname = field
-                data[ftype, fname] = np.empty(size, dtype="float64")
-                for chunk in chunks:
-                    for grid in chunk.objs:
+            ind = 0
+            for chunk in chunks:
+                for grid in chunk.objs:
+                    nd = 0
+                    for field in fields:
+                        ftype, fname = field
                         foffset = grid._index._field_offsets[fname]
                         values = self._read_single_field(fh, foffset)
-                        offset += grid.select(selector, values, data[field], offset)
+                        nd = grid.select(selector, values, data[field], ind)
+                    ind += nd
         return data
 
     def _read_particle_coords(self, chunks, ptf):


neutrinoceros:
Thanks a lot for working on this !
Honestly this method may be the one I understand the least, and a lot of writing a frontend is just mutating code that works from existing ones, so I don't think I can answer your question.

Anyway the important point is that it works, can you add a test as well ?

xshaokun:
> can you add a test as well ?

Sure, I'll write it tomorrow (in China). Should I modify the version as well?

neutrinoceros:
Take your time !
Yes, feel free to bump the patch number (note that it has to be done in `setup.cfg` and `yt_idefix/__init__.py`, I'll release right after merging :)

xshaokun:
I updated the PR summary. You can review it now ( •̀ᴗ•́ )و ̑̑

yt_idefix/fields.py
```diff
@@ -15,6 +17,19 @@ class IdefixVtkFieldInfo(FieldInfoContainer):
...
+        def _temperature(field, data):
+            mu = getattr(data.ds, "mu", compute_mu(data.ds.default_species_fields))
+            return mu * mh * data["gas", "pressure"] / data["gas", "density"] / kboltz
+
+        self.add_field(
+            name=("gas", "temperature"),
+            function=_temperature,
+            sampling_type="cell",
+            units=unit_system["temperature"],
+        )
```
neutrinoceros, line 31: This goes beyond the initial intention for this PR, and I'm not sure this definition is correct for Idefix data, so I'd prefer if we could make this definition Pluto-only for now, and preferably in a different PR.

neutrinoceros (dismissed):
So I see you're adding more stuff than just a single bugfix here. This would normally be fine, but here I'm finding that your tests break on my machine (even though they pass on GH !), for a reason I don't yet understand,
so I would like to request you keep this PR minimal and restricted to fix #67 only, so it's easier to get right. Of course you may open a separate PR with the field/units changes you're adding for Pluto.

yt_idefix/fields.py
xshaokun: OK. I'll retrieve this implement and create a new PlutoFieldInfo in the future when I prepare to.

neutrinoceros: thanks !

xshaokun:
> but here I'm finding that your tests break on my machine (even though they pass on GH !), for a reason I don't yet understand,

The tests also break on my machine if I use GH latest version of yt, but it's ok when I checkout to stable release (v4.0.1), for your reference.

neutrinoceros:
Ah, now I remember that I refactored the test framework here to allow me to store large datasets locally and including them in my test without having to add them to the repo. This is another reason why I'm seeing different results. That said I'm the tests are also failing (locally) for "sedov" and "sod" datasets (part of the repo) with any version of yt.
For now let's forget about the dev version of yt, I can fix that regression later, but we need to make sure everything pass with yt 4.0.1
I'm taking a break for now but I will come back to this later today.

xshaokun:
Take your time! btw, the tests are fine with yt 4.0.1 on my machine (MacOS 12.0.1). If there's anything needed to adapt or test on my side, I'm willing to help.

neutrinoceros (approved):
Ok so after a fresh install locally the I'm not getting failures locally anymore; I won't claim I understand though. I _also_ don't get any failures with the tip of yt's dev branch, so I'm tempted to accept as is.

I've checked that your test pass with the patch _and_ fails on the main branch, which is exactly what we want.

Now the only mystery remaining is why _you're_  still seeing some failures against the tip of yt. Can you share the traceback you're seeing ?

xshaokun:
This time they all passed with yt's dev branch. Before that I pulled all the commits I missed, which might solve my failures.

neutrinoceros:
Ok thanks for your feedback. I'm going to merge this and wait for the "bleeding edge" CI job to complete before I actually release it, but if all goes well I'll be only be a few minutes.
Thanks again for working on this, I really appreciate your contributions !

xshaokun:
You're wellcome! As a newcomer to yt, I'm getting more and more familiar with how to use yt and how it works during the process. 

I also appreciate this frontend you provide, which gives me avenue to use yt. Now I realized that creating a frontend is far more complicated than I thought. Thanks open source projects. I really enjoy it!

PR #69, opened 2022-01-17
ENH: enable to override all default_units with units_override (Pluto)

xshaokun:
### PR Summary

- enable to derive all default_units with the units passed to `units_override`
- adds test for `units_override`

### DIFF
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -1,4 +1,5 @@
 import os
+import re
 
 import pytest
 from more_itertools import distinct_combinations
@@ -61,27 +62,31 @@ def test_pluto_units(pluto_vtk_file):
 def test_pluto_complete_units_override(pluto_vtk_file):
     su = SAMPLE_UNITS
     unit_combs = distinct_combinations(su, 3)
-    invalid = PlutoVtkDataset.invalid_unit_combinations
     for comb in unit_combs:
         uo = {}
         for unit in comb:
             uo[unit] = su[unit]
-        if set(uo) not in invalid:
-            ds = yt_idefix.load(
-                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
-            )
-            assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
-            assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
-            assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
-            assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
-            assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
-            assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
+        if set(uo) in PlutoVtkDataset.invalid_unit_combinations:
+            continue
+        ds = yt_idefix.load(
+            pluto_vtk_file["path"],
+            geometry=pluto_vtk_file["geometry"],
+            units_override=uo,
+        )
+        assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
+        assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
+        assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
+        assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
+        assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
+        assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
 
 
 def test_pluto_one_unit_override(pluto_vtk_file):
     # Pluto's length_unit and density_unit will be combined with
     uo = {"time_unit": (2.0, "yr")}
-    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
+    ds = yt_idefix.load(
+        pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
+    )
     expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
     assert_allclose_units(ds.velocity_unit, expect_velocity)
 
@@ -89,7 +94,9 @@ def test_pluto_one_unit_override(pluto_vtk_file):
 def test_pluto_two_unit_override(pluto_vtk_file):
     # Pluto's length_unit will be combined with
     uo = {"time_unit": (2.0, "yr"), "density_unit": (32.0, "g/cm**3")}
-    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
+    ds = yt_idefix.load(
+        pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
+    )
     expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
     expect_mass = ds.quan(*uo["density_unit"]) * ds.length_unit ** 3
     assert_allclose_units(ds.velocity_unit, expect_velocity)
@@ -97,11 +104,12 @@ def test_pluto_two_unit_override(pluto_vtk_file):
 
 
 def test_pluto_invalid_units_override(pluto_vtk_file):
-    invalid = PlutoVtkDataset.invalid_unit_combinations
-    for uo in invalid:
+    for uo in PlutoVtkDataset.invalid_unit_combinations:
         with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
             yt_idefix.load(
-                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
+                pluto_vtk_file["path"],
+                geometry=pluto_vtk_file["geometry"],
+                units_override=uo,
             )
 
 
@@ -119,7 +127,9 @@ def test_pluto_temperature_unit_override(pluto_vtk_file):
         ),
     ):
         yt_idefix.load(
-            pluto_vtk_file["path"], geometry="cartesian", units_override=sample
+            pluto_vtk_file["path"],
+            geometry=pluto_vtk_file["geometry"],
+            units_override=sample,
         )
 
 
@@ -127,12 +137,16 @@ def test_pluto_over_units_override(pluto_vtk_file):
     with pytest.raises(
         ValueError,
         match=(
-            "More than 3 degrees of freedom were specified "
-            "in units_override \\(6 given\\)"
+            re.escape(
+                "More than 3 degrees of freedom were specified "
+                "in units_override (6 given)"
+            )
         ),
     ):
         yt_idefix.load(
-            pluto_vtk_file["path"], geometry="cartesian", units_override=SAMPLE_UNITS
+            pluto_vtk_file["path"],
+            geometry=pluto_vtk_file["geometry"],
+            units_override=SAMPLE_UNITS,
         )
 
 
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -62,8 +62,12 @@ def test_pluto_units_override(pluto_vtk_file):
         uo = {}
         for unit in comb:
             uo[unit] = ug[unit]
-
-        if set(uo) not in invalid:
+        if set(uo) in invalid:
+            with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
+                yt_idefix.load(
+                    file["path"], geometry=file["geometry"], units_override=uo
+                )
+        else:
             ds = yt_idefix.load(
                 file["path"], geometry=file["geometry"], units_override=uo
             )
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -11,6 +11,18 @@ from yt_idefix.api import IdefixVtkDataset, PlutoVtkDataset
 
 YT_VERSION = Version(yt.__version__)
 
+# A sample list of units for test.
+# The first three values are chosen randomly
+# and others are calculated correspondingly.
+SAMPLE_UNITS = {
+    "time_unit": (2.0, "s"),
+    "length_unit": (4.0, "cm"),
+    "mass_unit": (5.0, "kg"),
+    "density_unit": (0.078125, "kg/cm**3"),
+    "velocity_unit": (2.0, "cm/s"),
+    "magnetic_unit": (62.66570686577499, "gauss"),
+}
+
 
 def test_class_validation(vtk_file):
     file = vtk_file
@@ -46,34 +58,82 @@ def test_pluto_units(pluto_vtk_file):
         assert_allclose_units(getattr(ds, f"{u}_unit"), expected)
 
 
-def test_pluto_units_override(pluto_vtk_file):
-    file = pluto_vtk_file
-    ug = {
-        "time_unit": (2.0, "s"),
-        "length_unit": (4.0, "cm"),
-        "mass_unit": (5.0, "kg"),
-        "density_unit": (0.078125, "kg/cm**3"),
-        "velocity_unit": (2.0, "cm/s"),
-        "magnetic_unit": (62.66570686577499, "gauss"),
-    }
-    unit_combs = distinct_combinations(ug, 3)
-    invalid = yt_idefix.PlutoVtkDataset.invalid_unit_combinations
+def test_pluto_complete_units_override(pluto_vtk_file):
+    su = SAMPLE_UNITS
+    unit_combs = distinct_combinations(su, 3)
+    invalid = PlutoVtkDataset.invalid_unit_combinations
     for comb in unit_combs:
         uo = {}
         for unit in comb:
-            uo[unit] = ug[unit]
-        if set(uo) in invalid:
-            with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
-                yt_idefix.load(
-                    file["path"], geometry=file["geometry"], units_override=uo
-                )
-        else:
+            uo[unit] = su[unit]
+        if set(uo) not in invalid:
             ds = yt_idefix.load(
-                file["path"], geometry=file["geometry"], units_override=uo
+                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
             )
-            assert_allclose_units(ds.length_unit, ds.quan(*ug["length_unit"]))
-            assert_allclose_units(ds.mass_unit, ds.quan(*ug["mass_unit"]))
-            assert_allclose_units(ds.time_unit, ds.quan(*ug["time_unit"]))
+            assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
+            assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
+            assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
+            assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
+            assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
+            assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
+
+
+def test_pluto_one_unit_override(pluto_vtk_file):
+    # Pluto's length_unit and density_unit will be combined with
+    uo = {"time_unit": (2.0, "yr")}
+    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
+    expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
+    assert_allclose_units(ds.velocity_unit, expect_velocity)
+
+
+def test_pluto_two_unit_override(pluto_vtk_file):
+    # Pluto's length_unit will be combined with
+    uo = {"time_unit": (2.0, "yr"), "density_unit": (32.0, "g/cm**3")}
+    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
+    expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
+    expect_mass = ds.quan(*uo["density_unit"]) * ds.length_unit ** 3
+    assert_allclose_units(ds.velocity_unit, expect_velocity)
+    assert_allclose_units(ds.mass_unit, expect_mass)
+
+
+def test_pluto_invalid_units_override(pluto_vtk_file):
+    invalid = PlutoVtkDataset.invalid_unit_combinations
+    for uo in invalid:
+        with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
+            yt_idefix.load(
+                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
+            )
+
+
+def test_pluto_temperature_unit_override(pluto_vtk_file):
+    sample = {
+        "time_unit": (2.0, "s"),
+        "length_unit": (4.0, "cm"),
+        "temperature_unit": (2.0, "K"),
+    }
+    with pytest.raises(
+        ValueError,
+        match=(
+            "Temperature is not allowed in units_override, "
+            "since it's always in Kelvin in PLUTO"
+        ),
+    ):
+        yt_idefix.load(
+            pluto_vtk_file["path"], geometry="cartesian", units_override=sample
+        )
+
+
+def test_pluto_over_units_override(pluto_vtk_file):
+    with pytest.raises(
+        ValueError,
+        match=(
+            "More than 3 degrees of freedom were specified "
+            "in units_override \\(6 given\\)"
+        ),
+    ):
+        yt_idefix.load(
+            pluto_vtk_file["path"], geometry="cartesian", units_override=SAMPLE_UNITS
+        )
 
 
 def test_pluto_wrong_definitions_header(pluto_vtk_file):
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -53,8 +53,6 @@ def test_parse_pluto_metadata(pluto_vtk_file):
 def test_pluto_units(pluto_vtk_file):
     file = pluto_vtk_file
     ds = yt_idefix.load(file["path"])
-    if "units" not in file:
-        return
     for u, expected in file["units"].items():
         assert_allclose_units(getattr(ds, f"{u}_unit"), expected)
 
@@ -73,30 +71,30 @@ def test_pluto_complete_units_override(pluto_vtk_file):
             geometry=pluto_vtk_file["geometry"],
             units_override=uo,
         )
-        assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
-        assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
-        assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
-        assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
-        assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
-        assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
+        for attr, value in su.items():
+            assert_allclose_units(getattr(ds, attr), ds.quan(*value))
 
 
 def test_pluto_one_unit_override(pluto_vtk_file):
+    file = pluto_vtk_file
     # Pluto's length_unit and density_unit will be combined with
     uo = {"time_unit": (2.0, "yr")}
-    ds = yt_idefix.load(
-        pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
-    )
+    ds = yt_idefix.load(file["path"], geometry=file["geometry"], units_override=uo)
+    # Pluto's length_unit should be preserved
+    assert_allclose_units(ds.length_unit, file["units"]["length"])
+    # Pluto's velocity_unit should be overrided
     expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
     assert_allclose_units(ds.velocity_unit, expect_velocity)
 
 
-def test_pluto_two_unit_override(pluto_vtk_file):
+def test_pluto_two_units_override(pluto_vtk_file):
+    file = pluto_vtk_file
     # Pluto's length_unit will be combined with
     uo = {"time_unit": (2.0, "yr"), "density_unit": (32.0, "g/cm**3")}
-    ds = yt_idefix.load(
-        pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
-    )
+    ds = yt_idefix.load(file["path"], geometry=file["geometry"], units_override=uo)
+    # Pluto's length_unit should be preserved
+    assert_allclose_units(ds.length_unit, file["units"]["length"])
+    # Pluto's velocity_unit should be overrided
     expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
     expect_mass = ds.quan(*uo["density_unit"]) * ds.length_unit ** 3
     assert_allclose_units(ds.velocity_unit, expect_velocity)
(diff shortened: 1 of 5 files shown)


yt_idefix/data_structures.py
```diff
@@ -427,15 +427,45 @@ def _get_time(self):
...
 
         self.mass_unit = density_unit * self.length_unit ** 3
         self.time_unit = self.length_unit / self.velocity_unit
         self.magnetic_unit = np.sqrt(4.0 * np.pi * density_unit) * self.velocity_unit
         self.magnetic_unit.convert_to_units("gauss")
+
+    default_units = {
+        "length_unit": "cm",
+        "time_unit": "s",
+        "density_unit": "g/cm**3",
```
neutrinoceros, line 449: please add a comment here to note that `"density_unit"` is the one key that's actually not in the base `Dataset` class

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,45 @@ def _get_time(self):
...
+        # other combinations might be added in the future
+        valid_keys = {"length_unit", "velocity_unit", "density_unit"}
+        invalid_keys_found = set(units_override.keys()) - valid_keys
+        if invalid_keys_found:
+            raise ValueError(
+                f"units_override contains invalid keys: {invalid_keys_found}. "
+                f"Only allow {valid_keys} to be passed to units_override"
+            )
+
+        super(cls, cls)._validate_units_override_keys(units_override)
```
neutrinoceros, line 471: Ah so if I'm not mistaken, you don't actually need to override this method, as you would get the same result by just changing `IdefixVtkDataset.default_units` to only have the three keys you consider valid here `("length_unit", "velocity_unit", "density_unit")`. However, this is probably not a valid approach because then we wouldn't have a default for magnetic and temperature units.

Most importantly, your current implementation is problematic because it prevents using otherwise valid unit normalisation (in `mass_unit` should still be usable).

I believe I designed (or refactored, rather) this class method in yt a while ago. I believe my intention was to allow _extending_ the list of valid units overrides, but it's probably not a good idea to restrict it, compared to the base class.

neutrinoceros (changes requested):
While I can certainly get behind what you want to enable here, I think the current implementation isn't backward compatible, but we can certainly evolve it to be !

Also I think this count as a feature rather than a bugfix, so you could just bump the version number to 0.10.0

yt_idefix/data_structures.py
xshaokun: Your point is right. I found the method you designed in the AMRVAC frontend, that's why I commented here that other combinations would be supported in the future. 

I planed to make units_override usable for now. Maybe I don't have to rush to create this PR, so I'm considerring putting this PR on hold until I make this to be a mature feature.

xshaokun: My intention is to reduce the dependece on header file `definitions.h` during analysis, and be able to scale the results to other physical system by adapting normalization factor in post-analysis.

neutrinoceros: >  I'm considerring putting this PR on hold until I make this to be a mature feature.

That is fine by me.

> My intention is to reduce the dependece on header file definitions.h during analysis, and be able to scale the results to other physical system by adapting normalization factor in post-analysis.

Sounds like a reasonable approach.

xshaokun:
Oops I have a difficulty now and would like to share some thoughts with you. If you have any advice, that would be great.

the `default_unit` now consists seven units: length, time, velocity, density, magnetic and temperatuer. In Pluto, [length, density, velocity] are used as base units and are used to derived other units. But in post-analysis, beside temperature should not be override since it's always in Kelvin in Pluto, there's no such limitation and users are able to derive all units based on any combinations of three units other than [length, velocity, time]. 

The difficulty is there are too much combinations! The worst method is to use many if/else statements to derive other units based on different combinations, but that must be terrible. So I'm trying to find some constraints to reduce valid combinations.

neutrinoceros:
> The difficulty is there are too much combinations! The worst method is to use many if/else statements to derive other units based on different combinations, but that must be terrible. So I'm trying to find some constraints to reduce valid combinations.

yeah that sounds difficult to get right and maintain. I am not sure how I would approach this but I would advise you consider generating the list of valid combinations instead of hard-coding it. I'm thinking it could be relatively easy using `itertools` functions or possibly `more_itertools` (third party, but already a hard dependency to yt it self, so a transitive dependency to yt_idefix).

xshaokun:
> I'm thinking it could be relatively easy using itertools functions or possibly more_itertools

Thanks! I”ll take a look tomorrow.

xshaokun:
Many condition statements cannot be avoided for now, but it's much better than I expected. I'll come back tomorrow if you have any comments :)

neutrinoceros:
I cannot promise to review this before next Wednesday, but I'll try.

xshaokun:
It's okay，take your time

tests/test_vtk.py
```diff
@@ -45,6 +46,36 @@ def test_pluto_units(pluto_vtk_file):
...
+    ug = {
+        "time_unit": (2.0, "s"),
+        "length_unit": (4.0, "cm"),
+        "mass_unit": (5.0, "kg"),
+        "density_unit": (0.078125, "kg/cm**3"),
+        "velocity_unit": (2.0, "cm/s"),
+        "magnetic_unit": (62.66570686577499, "gauss"),
+    }
+    unit_combs = distinct_combinations(ug, 3)
+    invalid = yt_idefix.PlutoVtkDataset.invalid_unit_combinations
```
neutrinoceros, line 60: please import the  PlutoVtkDataset class at the top level. It'll be easier to migrate this frontend to yt's main repo if every name changes are gathered at the top level of each files (or as much as possible)
```suggestion
    invalid = PlutoVtkDataset.invalid_unit_combinations
```

tests/test_vtk.py
```diff
@@ -45,6 +46,36 @@ def test_pluto_units(pluto_vtk_file):
...
+def test_pluto_units_override(pluto_vtk_file):
+    file = pluto_vtk_file
+    ug = {
+        "time_unit": (2.0, "s"),
+        "length_unit": (4.0, "cm"),
+        "mass_unit": (5.0, "kg"),
+        "density_unit": (0.078125, "kg/cm**3"),
+        "velocity_unit": (2.0, "cm/s"),
+        "magnetic_unit": (62.66570686577499, "gauss"),
+    }
```
neutrinoceros, line 58: Can you add a comment to explain how you chose these values (randomly ? or is this a specific set of consistent values ?)

tests/test_vtk.py
```diff
@@ -45,6 +46,36 @@ def test_pluto_units(pluto_vtk_file):
...
+    unit_combs = distinct_combinations(ug, 3)
+    invalid = yt_idefix.PlutoVtkDataset.invalid_unit_combinations
+    for comb in unit_combs:
+        uo = {}
+        for unit in comb:
+            uo[unit] = ug[unit]
+        if set(uo) in invalid:
+            with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
```
neutrinoceros, line 66: I'd like the error message to be even clearer. Please explicitly say that the input units_override isn't valid

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
     def _set_code_unit_attributes(self):
         """Conversion between physical units and code units."""
-        length_unit = self.parameters.get("unit_length", 1.0)
-        velocity_unit = self.parameters.get("unit_velocity", 1.0)
-        density_unit = self.parameters.get("unit_density", 1.0)
 
-        self.length_unit = self.quan(length_unit, "cm")
-        self.velocity_unit = self.quan(velocity_unit, "cm/s")
-        density_unit = self.quan(density_unit, "g/cm**3")
+        # Overriding units are preferred to PLUTO's base units
```
neutrinoceros, line 431:
```suggestion
        # User-passed units are preferred to PLUTO's base units
```

neutrinoceros (commented):
I didn't have time to look at everything but here are a couple comments already. I'll finish reading this soon

tests/test_vtk.py
xshaokun: It makes sense!

tests/test_vtk.py
xshaokun: Sure. I just chose the first three values arbitraily and derived others to make them consistent.

tests/test_vtk.py
xshaokun: Okay, I'll improve it

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+            )
+
+        # check if provided overrides are allowed
+        suo = set(units_override)
+        if suo in cls.invalid_unit_combinations:
+            raise ValueError(
+                f"Combination {suo} passed to units_override "
+                "cannot derive all units\n"
+                f"Choose any other combinations, except for:\n {cls.invalid_unit_combinations}"
+            )
```
xshaokun, line 599: @neutrinoceros Since there are too many valid combinations, I only listed invalid combinations here to let users avoid. Should I explicitly show this verbose message in the test?

neutrinoceros: You approach seems reasonable, but yes, please include a test to coverage this path too.

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+        # are valid other than the three cases listed explicitly.
+
+        if "temperature_unit" in units_override:
+            raise ValueError(
+                "Temperature is not allowed in units_override,"
+                "since it's always in Kelvin in PLUTO"
+            )
+
+        # Three units are enough for deriving others, more will cause conflict
+        if len(units_override) > 3:
```
neutrinoceros, line 586: Are you sure that's correct ? I would assume that overriding magnetic as well as length, mass and time is fine ?

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+
+        # YT supports overriding other normalisations, this method ensures consistency
+        # between supplied 'units_override' items and principles in PLUTO.
+
+        # PLUTO's normalisations/units have 3 degrees of freedom. Therefore, any combinations
+        # are valid other than the three cases listed explicitly.
+
+        if "temperature_unit" in units_override:
+            raise ValueError(
+                "Temperature is not allowed in units_override,"
```
neutrinoceros, line 581: I'm not 100% sure this is fully consistent with your inline comment above that states that user-passed overrides are preferred over Pluto's internal units. I could easily be convinced that it's fine, however.

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
-        density_unit = self.quan(density_unit, "g/cm**3")
+        # Overriding units are preferred to PLUTO's base units
+        pluto_units = {
+            "length_unit": "cm",
+            "velocity_unit": "cm/s",
+            "density_unit": "g/cm**3",
+        }
+        spu = set(pluto_units)
+        if len(self.units_override) < 3:
+            if len(self.units_override):
```
neutrinoceros, line 439: equivalent but clearer (in my opinion)
```suggestion
            if len(self.units_override) > 0:
```

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+                ytLogger.info(
+                    "Less than 3 units were specified in units_override (%s). "
```
neutrinoceros, line 441:
```suggestion
                    "Less than 3 units were specified in units_override (got %s). "
```

yt_idefix/data_structures.py
xshaokun: Since we have only three base units here: length, mass and time, just like solving a system of equations containing three unknowns, three equations are enough, or three base/compound units here.

So an additional magnetic might conflict with that derived with length, mass and time. "more will _likely_ cause conflict" is more precise I think.

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+                    "Need to rely on PLUTO's internal units to derive other units",
+                    len(self.units_override),
+                )
+
+            spu.difference_update(set(self.units_override))
```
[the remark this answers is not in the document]
xshaokun: > note however that I don't think `self.units_override` should be updated internally. Because this attribute is user-facing, it should only reflect exactly what the user passed at load time, so I'd advise using a different intermediate variable here

I did considered that! But I also want to let users know from which three units the others are derived when they only pass 1 or 2 units, even though the additional units picked from Pluto definitions don't be overrided really. Maybe I should create another attribute?

xshaokun:
I think I should explain more about my strategy.

The default case is no value paased to `units_override`, then the three units (density, velocity, length) from Pluto's definition are used to derived others.

In Pluto, units of density, velocity and length are unity if user doesb't specify. The `units_override` should be used when users wanted to override the (Pluto's) default units.

If the user only passes 1 unit (or two) which is not able to derive othes, so it requires two additional units. Obviously We should choose default units from PLUTO's definiton which they are consistent with valid combination.

neutrinoceros:
Okay that makes sense. Can you also explain how magnetic units are derived here ? I don't understand how they are computed from length time and mass but without any electric charge unit.

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,175 @@ def _get_time(self):
...
+        self.mass_unit = uo["mass_unit"]
+        self.length_unit = uo["length_unit"]
+        self.time_unit = uo["time_unit"]
+
+        self.velocity_unit = self.length_unit / self.time_unit
+        self.density_unit = self.mass_unit / self.length_unit ** 3
+        self.magnetic_unit = (
+            np.sqrt(4.0 * np.pi * self.density_unit) * self.velocity_unit
+        )
         self.magnetic_unit.convert_to_units("gauss")
```
xshaokun, line 546: @neutrinoceros Here's formula to derive magnetic, where density and velocity can be replaced with mass, time and length. There's no need of charge here because in cgs it's defined by mass, time and length

yt_idefix/data_structures.py
xshaokun: It's already covered in the test.

tests/test_vtk.py
xshaokun: Here I just wanted to match the error messeage to make sure it's the `ValueError` I expected, not caused by something else. I think the regular expression is enough, right?

yt_idefix/data_structures.py
xshaokun: Honestly speaking, I didn't derive the formula by myself but it's used in Pluto and also in other frontends such as athena and amrvac.

xshaokun: In cgs, permittivity = permeability = 1 (dimensionless)

neutrinoceros: Fair enough. I'll admit I've never sat through this derivation so I don't have a grasp of what "cgs" really means in the MHD simulation realm. I'll just trust you here as I did with other collaborators and frontends :)

neutrinoceros:
I'm planning to review this fully tomorrow unless you still want to pour more changes before I do.

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+        if set(uo) not in invalid:
+            ds = yt_idefix.load(
+                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
+            )
+            assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
+            assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
+            assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
+            assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
+            assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
+            assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
```
neutrinoceros, line 78:
```suggestion
    for comb in unit_combs:
        uo = {}
        for unit in comb:
            uo[unit] = su[unit]
        if set(uo) in PlutoVtkDataset.invalid_unit_combinations:
            return
        ds = yt_idefix.load(
            pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
        )
        assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
        assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
        assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
        assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
        assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
        assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
```

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+
+
+def test_pluto_one_unit_override(pluto_vtk_file):
+    # Pluto's length_unit and density_unit will be combined with
+    uo = {"time_unit": (2.0, "yr")}
+    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
```
neutrinoceros, line 84:
```suggestion
    ds = yt_idefix.load(pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo)
```

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+    assert_allclose_units(ds.velocity_unit, expect_velocity)
+    assert_allclose_units(ds.mass_unit, expect_mass)
+
+
+def test_pluto_invalid_units_override(pluto_vtk_file):
+    invalid = PlutoVtkDataset.invalid_unit_combinations
+    for uo in invalid:
+        with pytest.raises(ValueError, match=r".* cannot derive all units\n.*"):
+            yt_idefix.load(
+                pluto_vtk_file["path"], geometry="cartesian", units_override=uo
```
neutrinoceros, line 104:
```suggestion
                pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
```

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+    uo = {"time_unit": (2.0, "yr")}
+    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
+    expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
+    assert_allclose_units(ds.velocity_unit, expect_velocity)
+
+
+def test_pluto_two_unit_override(pluto_vtk_file):
+    # Pluto's length_unit will be combined with
+    uo = {"time_unit": (2.0, "yr"), "density_unit": (32.0, "g/cm**3")}
+    ds = yt_idefix.load(pluto_vtk_file["path"], geometry="cartesian", units_override=uo)
```
neutrinoceros, line 92:
```suggestion
    ds = yt_idefix.load(pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo)
```

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+            pluto_vtk_file["path"], geometry="cartesian", units_override=sample
+        )
+
+
+def test_pluto_over_units_override(pluto_vtk_file):
+    with pytest.raises(
+        ValueError,
+        match=(
+            "More than 3 degrees of freedom were specified "
+            "in units_override \\(6 given\\)"
```
neutrinoceros, line 131: please avoid complicated escape sequences (`\\`) and use `re.escape` with raw strings instead.

tests/test_vtk.py
```diff
@@ -45,6 +58,84 @@ def test_pluto_units(pluto_vtk_file):
...
+        ),
+    ):
+        yt_idefix.load(
+            pluto_vtk_file["path"], geometry="cartesian", units_override=SAMPLE_UNITS
```
neutrinoceros, line 135:
```suggestion
            pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=SAMPLE_UNITS
```

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
     def _set_code_unit_attributes(self):
         """Conversion between physical units and code units."""
-        length_unit = self.parameters.get("unit_length", 1.0)
-        velocity_unit = self.parameters.get("unit_velocity", 1.0)
-        density_unit = self.parameters.get("unit_density", 1.0)
 
-        self.length_unit = self.quan(length_unit, "cm")
-        self.velocity_unit = self.quan(velocity_unit, "cm/s")
-        density_unit = self.quan(density_unit, "g/cm**3")
+        # User-passed units are preferred to PLUTO's base units for deriving all units
```
neutrinoceros, line 431:
```suggestion
        # User-passed units are preferred over PLUTO's base units for deriving all units
```

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
+        spu = set(pluto_units)
+        uo = self.units_override.copy()
+        if len(uo) < 3:
+            if len(uo) > 0:
+                ytLogger.info(
+                    "Less than 3 units were specified in units_override (got %s). "
+                    "Need to rely on PLUTO's internal units to derive other units",
+                    len(uo),
+                )
+            # Remove those already in units_override
```
neutrinoceros, line 446: for extra clarity
```suggestion
            # Remove those already in self.units_override
```

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
+            spu.difference_update(uo)
+            for unit in spu:
```
neutrinoceros, line 448: this loop looks a lot like a `while len(uo) < 3`. In any case I'm pretty sure it's more complex than it needs to be, so please try to simplify this a bit.

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
+                        self._validate_units_override_keys(uo)
+                        ytLogger.info("Relying on %s: %s.", unit, uo[unit])
+                    except ValueError:
+                        del uo[unit]
+                if len(uo) == 3:
+                    break
+
+        # Then we calculate base units via the units in units_override
+        # The condition statements are essential to prevent getting stuck in infinite recursion
+        # However they are fragile, you must be cautious to change them
```
neutrinoceros, line 461:
```suggestion
        # However they are fragile, one must be cautious to change them
```

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
+        def _time():
```
neutrinoceros, line 462: The ways these functions are designed is not immediately obvious to me. I guess I'd prefer having them named like "_setup_time_unit" ? Also the fact that they have only side-effects (they modify a variable from the outer scope rather than returning values) is a little unsettling. Can you please add some more comments to explain your strategy ?

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,173 @@ def _get_time(self):
...
+            if "time_unit" not in uo:
+                uo["time_unit"] = uo["length_unit"] / uo["velocity_unit"]
```
neutrinoceros, line 464: I guess we could lower the indentation level by using guard conditions instead, such as
```suggestion
            if "time_unit" in uo:
                return
            uo["time_unit"] = uo["length_unit"] / uo["velocity_unit"]
```

neutrinoceros (commented):
Alright I've read everything and I have two overarching remarks:
- I'd like to see some simplifications done before I'm confortable merging this (remember that one must be twice as smart to debug code than to write it, and this is pretty clever code already ;) )
- I'm really hoping I can count on your help and guidance in the future if we find new bugs here.

In short, I'm okay with everything you're adding here, but I'm not yet confortable maintaining it on my own. I'm sure we can make a couple things a little simpler still :)

yt_idefix/data_structures.py
xshaokun: Thanks! Just forgot `while`...

yt_idefix/data_structures.py
xshaokun: Sure.

xshaokun: > Can you please add some more comments to explain your strategy ?

Do you mean the strategy why did I set such condition statements?

I just tried different conditions during the debugging. So I'm sorry to tell you that I don't have a straightforward strategy to constrcut this part. But each one of them is useful if you test it with all the valid combinations.

xshaokun:
Thanks for your reviewing! I did some improvements as your suggestions.

Sometimes I just don't know which kind of code is better, but I'm still learning:)

> I'm really hoping I can count on your help and guidance in the future if we find new bugs here.

Sure, I'm glad to!

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,199 @@ def _get_time(self):
...
+                    self._validate_units_override_keys(uo_cache)
+                    ytLogger.info("Relying on %s: %s.", unit, uo_cache[unit])
+                except ValueError:
+                    # It means the combination is invalid
+                    del uo_cache[unit]
+
+        # Some unit_setup_functions
+        # The condition statements are essential to prevent getting stuck in infinite recursion
+        # However they are fragile, one must be cautious to change them
+        def _setup_time_unit(cache):
```
neutrinoceros, line 476: so it looks like we slightly misunderstood each other (no big deal !). I suggested to rename these functions as `_setup_<x>_unit` when they were designed around side effects on an outer-scope dictionary. Now that you evolved the design to return actual values instead, a more appropriate naming convention should be `_get_<x>_unit`. I'm sorry if my original comment caused confusion.

I am not opposing your original design btw. It's fine to use side effects if you want, especially in nested functions like these.

I'm thinking however this setup function, and the myriad of nested functions you defined, look more and more like a separate, tiny class with a very specific responsibility. What do you think ?

yt_idefix/data_structures.py
neutrinoceros: Oh, my bad, thanks !

neutrinoceros (commented):
Another round of comments. I'll aim to do a final review by Friday, and most likely merge + release then. Thanks again for your continued effort !

tests/test_vtk.py
```diff
@@ -45,6 +59,97 @@ def test_pluto_units(pluto_vtk_file):
...
+            pluto_vtk_file["path"],
+            geometry=pluto_vtk_file["geometry"],
+            units_override=uo,
+        )
+        assert_allclose_units(ds.length_unit, ds.quan(*su["length_unit"]))
+        assert_allclose_units(ds.mass_unit, ds.quan(*su["mass_unit"]))
+        assert_allclose_units(ds.time_unit, ds.quan(*su["time_unit"]))
+        assert_allclose_units(ds.density_unit, ds.quan(*su["density_unit"]))
+        assert_allclose_units(ds.velocity_unit, ds.quan(*su["velocity_unit"]))
+        assert_allclose_units(ds.magnetic_unit, ds.quan(*su["magnetic_unit"]))
```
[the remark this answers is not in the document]
xshaokun: Of course! How could I forget it!

yt_idefix/data_structures.py
xshaokun: Oh it’s my first time seeing the term “side effects” in programming, I thought you mean it's bad. Now I know its real meaning.

Due to their similar function names, I wanted to use `eval` instead of a dictionary here. But I remembered you said that we should avoid using `eval`.

> 'm thinking however this setup function, and the myriad of nested functions you defined, look more and more like a separate, tiny class with a very specific responsibility. What do you think ?

For me, it's like a `unyt.UnitSystem` which however only accepts mass, length, and time. So if you recommend, I'm willing to separate this part and wrap it into a class tomorrow. I’ve been thinking it's a little bit too much to define it in this method.

neutrinoceros: Sure. I'm not familiar with how extensible unyt.UnitSystem is but if you think it's appropriate you could maybe subclass it. Otherwise just move most of this function into a (private) class and call it a day :)

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,199 @@ def _get_time(self):
...
+            uo_cache[unit] = value
+            # Make sure the combination of units are able to derive base units
+            # If self.units_override is empty, skip this part
+            if uo_size > 0:
+                try:
+                    self._validate_units_override_keys(uo_cache)
+                    ytLogger.info("Relying on %s: %s.", unit, uo_cache[unit])
+                except ValueError:
+                    # It means the combination is invalid
+                    del uo_cache[unit]
```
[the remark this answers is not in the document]
neutrinoceros: We'll ultimately you're the only user I know of for now so if you don't mind having them as info it's really up to you. I'm also fine with changing the level in a future PR too if you change your mind :)

tests/test_vtk.py
```diff
@@ -45,6 +59,97 @@ def test_pluto_units(pluto_vtk_file):
...
+
+
+def test_pluto_one_unit_override(pluto_vtk_file):
+    # Pluto's length_unit and density_unit will be combined with
+    uo = {"time_unit": (2.0, "yr")}
+    ds = yt_idefix.load(
+        pluto_vtk_file["path"], geometry=pluto_vtk_file["geometry"], units_override=uo
+    )
+    expect_velocity = ds.length_unit / ds.quan(*uo["time_unit"])
```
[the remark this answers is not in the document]
xshaokun:
```
expect_velocity = plut_vtk_file["units"]["length"] / ds.quan(*uo["time_unit"])
```
Maybe this is better. My intention is to check if right Pluto's unit was preserved and participated in unit derivation. The value of `expect_velocity` is different for each dataset because they have different defined units. It also depends on the unit in `uo` because if we give `length_unit` in `uo`, `density_unit` and `velocity_unit` will be preserved instead of `density_unit` and `length_unit` here. So I don't think it's a good idea to hardcode the expected value in `meta.yaml`.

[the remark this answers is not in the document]
neutrinoceros: Ok that makes sense

yt_idefix/data_structures.py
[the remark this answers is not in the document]
xshaokun: I'm really hoping there'll be more Pluto's user using this frontend in the future haha.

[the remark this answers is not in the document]
neutrinoceros: Of course, me too !

yt_idefix/data_structures.py
```diff
@@ -427,15 +427,217 @@ def _get_time(self):
...
+        if suo in cls.invalid_unit_combinations:
+            raise ValueError(
+                f"Combination {suo} passed to units_override "
+                "cannot derive all units\n"
+                f"Choose any other combinations, except for:\n {cls.invalid_unit_combinations}"
+            )
+
+        super(cls, cls)._validate_units_override_keys(units_override)
+
+    class __BaseUnits:
```
neutrinoceros, line 542: I'm surprised you wrote this as a _nested_ class. I'd much rather have this class be seperated, otherwise I don't see the point. Also, a double trailing underscore is not a convention I'm familiar with. I think a single one is enough to indicate the class is private, isn't it ?

xshaokun: Oh when I started writing this class, I wanted to include deriving other units as well. So it shouldn't work for other code and I considered it as a nested class. Now it becomes more general but I forgot to seperate it.

Do you think I should leave this class in this file, or move it to `definitions.py` and import it here? I prefer the latter.

neutrinoceros: It doesn't need to be usable anywhere else to be separated. Otherwise you can leave it in the same file or move it to definitions if you think that's cleaner

xshaokun: Well I decide to leave this work to the future when necessary.

xshaokun:
I'll push a new commit if you have no more suggustion :)

neutrinoceros:
Go ahead. I'm not sure we've settled the nested-class thread though.

xshaokun:
> I'm not sure we've settled the nested-class thread though.

I think there are no more questions (except for API for validation) on my side for now. I’ll come back tomorrow if you have any feedback.

neutrinoceros (approved):
Alright I see a couple minor nits I want to fix still but I'll do that myself in a follow up PR and will release in a couple hours.
Thanks again for your work, I really appreciate it !

xshaokun:
I was going to squash my commits before merging. Anyway thanks for your patient reviews, next time I'll do it better !

neutrinoceros:
No problem. I could have squashed on merge too but forgot to, it's no big deal :)

Issue #83, opened 2022-02-02
TODO: parse units and input parameters in unit definitions (Pluto)


xshaokun:
For the part of parsing units, I considered the case where defined constants are used to define units. However when I reviewed the test problems provided by PLUTO, I found units can be defined in such way:

``` 
#define  UNIT_LENGTH                    (5.2*CONST_au)
#define  UNIT_DENSITY                   (CONST_Msun/(UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH))
#define  UNIT_VELOCITY                  (sqrt(CONST_G*g_inputParam[Mstar]*CONST_Msun/UNIT_LENGTH)/(2.*CONST_PI))
```

where involves units defined previously and input parameters. So the implementation of parsing units should be improved or refactored in the future so that it can handle such definition.

neutrinoceros:
Thanks for reporting this ! For closure, any one else who would like to give this a try is welcome to.
(https://www.redhat.com/en/blog/dont-lick-cookie ;) )

neutrinoceros:
@xshaokun are you still interested in doing this ?
I'm thinking about opening a PR to merge yt_idefix into mainline yt, so I'd rather have any important change happen here first if possible.

xshaokun:
> @xshaokun are you still interested in doing this ?

Oh yes. I'm glad to know that this package will be merged into yt. 

I'll try to propose a solution for this issue in the following week.

neutrinoceros:
Thank for your quick reply. For the record, merging this into yt is still on the time scale of a couple month, so I mean no rush !

PR #84, opened 2022-02-02
TST: add Pluto datasets in polar and spherical coordinate

xshaokun:
Two new Pluto datasets:

- pluto_disk_planet: 3D HD disk-planet interaction problem in spherical coordinate
- pluto_rotor: 2D MHD rotor problem in polar coordinate

`test_pluto_wrong_definitions_header_with_geometry()` was modified only for making the datasets pass the test, while this test will be removed in future commit.

Due to #83 , the unit definitons in pluto_disk_planet are given by exact value for now.


tests/data/pluto_rotor/vtk.out
```diff
@@ -0,0 +1,31 @@
+0 0.000000e+00 1.000000e-06 0 single_file little rho vx1 vx2 vx3 Bx1 Bx2 Bx3 prs
```
neutrinoceros, line 1: Just curious, does the "little" column refer to endianness ? If so, we may have a bug to fix for the case big-endians are used. Idefix _always_ outputs in little endian so it's currently hard-coded in the frontend

neutrinoceros (approved):
Just a quick question that's not directly linked to this particular PR.

Thanks a lot for doing this, it'll be extremely valuable for my refactor PR !

tests/data/pluto_rotor/vtk.out
xshaokun: Yes, it refers to endianness.

xshaokun: But I’ve never changed the endianness before, so let me check it tomorrow to figure out how to specify it in PLUTO.

neutrinoceros: Sure, I'll open an issue so we don't loose this thread !

PR #90, opened 2022-02-03
BUG: fix incorrect conversion between code_time and cgs_unit

xshaokun:
The behavior of this bug can be shown by the following code:
``` Python
ds = yt.load("tests/data/pluto_disk_planet/data.0010.vtk")
print("current_time: ", ds.current_time)
print("current_time_cgs: ", ds.current_time.in_cgs())
print("time_unit: ", ds.time_unit)
codet = ds.quan(1, "code_time")
print("code_time", ds.unit_registry["code_time"])
print("codet: ", codet)
print("codet_cgs: ", codet.in_cgs())
```

The ouput is 
```
current_time:  1.0 code_time
current_time_cgs:  1.0 s
time_unit:  373173289.848891 s
code_time (373173289.848891, (time), 0.0, '\\rm{code\\ time}', False)
codet:  1 code_time
codet_cgs:  1.0 s
```

The reason is our too early calling of "code_time" in the definition of `self.current_time`, which is called before `self.set_units()`. So the conversion factor between code_unit and cgs_time hasn't setup yet. 

While I don't know why the definition of `code_time` is correct but is wrong in `ds.quan()` (see the behavior of `codet`)

In this PR, I just left `self.current_time` to be an int/float object and it is converted to unyt_quantity by `self._set_derived_attrs()` later.

### DIFF
--- a/tests/data/pluto_disk_planet/meta.yaml
+++ b/tests/data/pluto_disk_planet/meta.yaml
@@ -5,6 +5,7 @@ attrs:
   geometry: spherical
   dimensionality: 3
   kind: pluto
+  def_units: True
   units:
     length: "7.77909038e13 cm"
     velocity: "208457.8557 cm/s"
--- a/tests/data/pluto_disk_planet/meta.yaml
+++ b/tests/data/pluto_disk_planet/meta.yaml
@@ -5,7 +5,7 @@ attrs:
   geometry: spherical
   dimensionality: 3
   kind: pluto
-  def_units: True
+  has_units: True
   units:
     length: "7.77909038e13 cm"
     velocity: "208457.8557 cm/s"
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -293,7 +293,7 @@ class IdefixVtkDataset(IdefixDataset):
         self.domain_right_edge = dre
 
         # time wasn't stored in vtk files before Idefix 0.8
-        self.current_time = md.get("time", -1)
+        self.current_time = self.quan(md.get("time", -1), "code_time")
 
         # periodicity was not stored in vtk files before Idefix 0.9
         self._periodicity = md.get("periodicity", (True, True, True))
@@ -344,7 +344,7 @@ class IdefixDmpDataset(IdefixDataset):
             [fdata[f"xr{idir}"][-1] for idir in "123"], dtype="float64"
         )
 
-        self.current_time = fdata["time"]
+        self.current_time = self.quan(fdata["time"], "code_time")
 
         self._periodicity = tuple(bool(p) for p in fdata["periodicity"])
 
@@ -441,14 +441,16 @@ class PlutoVtkDataset(IdefixVtkDataset):
             )
         index = int(match.group(1))
 
-        self.current_time = -1
+        self.current_time = self.quan(-1, "code_time")
         if os.path.isfile(log_file):
             log_regexp = re.compile(rf"^{index}\s(\S+)")
             with open(log_file) as fh:
                 for line in fh.readlines():
                     log_match = re.search(log_regexp, line)
                     if log_match:
-                        self.current_time = float(log_match.group(1))
+                        self.current_time = self.quan(
+                            float(log_match.group(1)), "code_time"
+                        )
                         break
                 else:
                     ytLogger.warning(
@@ -457,6 +459,18 @@ class PlutoVtkDataset(IdefixVtkDataset):
                     )
         else:
             ytLogger.warning("Missing log file %s, setting current_time = -1", log_file)
+
+    def set_code_units(self):
+        super().set_code_units()
+        # Here we need to clean the unit cache
+        # to remove any cached code units with default conversion factor
+        self.unit_registry._unit_object_cache.clear()
+
+    def _set_derived_attrs(self):
+        super()._set_derived_attrs()
+        # Some attrs were initialized with code units but with default conversion factor,
+        # we need to update them after the code units are modified in self.set_code_units()
+        self.current_time = self.quan(self.current_time, "code_time")
 
     def _set_code_unit_attributes(self):
         """Conversion between physical units and code units."""
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -293,7 +293,7 @@ class IdefixVtkDataset(IdefixDataset):
         self.domain_right_edge = dre
 
         # time wasn't stored in vtk files before Idefix 0.8
-        self.current_time = self.quan(md.get("time", -1), "code_time")
+        self.current_time = md.get("time", -1)
 
         # periodicity was not stored in vtk files before Idefix 0.9
         self._periodicity = md.get("periodicity", (True, True, True))
@@ -344,7 +344,7 @@ class IdefixDmpDataset(IdefixDataset):
             [fdata[f"xr{idir}"][-1] for idir in "123"], dtype="float64"
         )
 
-        self.current_time = self.quan(fdata["time"], "code_time")
+        self.current_time = fdata["time"]
 
         self._periodicity = tuple(bool(p) for p in fdata["periodicity"])
 
@@ -441,16 +441,15 @@ class PlutoVtkDataset(IdefixVtkDataset):
             )
         index = int(match.group(1))
 
-        self.current_time = self.quan(-1, "code_time")
+        # will be converted to actual unyt_quantity in _set_derived_attrs
+        self.current_time = -1
         if os.path.isfile(log_file):
             log_regexp = re.compile(rf"^{index}\s(\S+)")
             with open(log_file) as fh:
                 for line in fh.readlines():
                     log_match = re.search(log_regexp, line)
                     if log_match:
-                        self.current_time = self.quan(
-                            float(log_match.group(1)), "code_time"
-                        )
+                        self.current_time = float(log_match.group(1))
                         break
                 else:
                     ytLogger.warning(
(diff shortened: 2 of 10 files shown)


neutrinoceros (changes requested):
On my phone right now but I see a problem with this approach: when typing contraint become stricter with a future version of yt, we shouldn't be able to change types of arbitrary attributes mid initialization. Is there a way we could use a placeholder unyt_quantity as a default value instead of an int (-1) ?

xshaokun:
Oh I think I probably misled you. `self._set_derived_attrs()` is a method of `Dataset` class, where yt has already implemented the type conversion of `self.curent_time`. It is the strategy provided by yt and also used by some of other frontends I saw, leaving the current_time to be an int or a float at first, so I think it won't change in the future, right?

> Is there a way we could use a placeholder unyt_quantity as a default value instead of an int (-1) ?

I think we can use `self.quan(-1, "s")` whose unit is wrong but it wouldn't affect dataset's unit_registry. Because unyt_quantity requires a scalar and a unit as input. If an additional `unit_registry` is specified, the unit would be used from or created in this registry. And `self.quan` uses the `self.unit_registry` in default.

neutrinoceros:
> frontends I noticed, leaving the current_time to be an int or a float at first, so I think it won't change in the future, right?

I am actively planning to change that in the future myself, unless it turns out to be impossible :)

> I think we can use self.quan(-1, "s")

Sounds like a better placeholder to me.

Ok so I don't think we can avoid adding a test here. This seems like a very delicate part to get right and there are probably many ways to get it wrong as in the current implementation.
I guess you could just adapt your script into a test and call it a day.

xshaokun:
> I am actively planning to change that in the future myself, unless it turns out to be impossible :)

Ok it makes sense. Shall I set the default value as `self.quan(-1, "s")` right now? While I prefer to do it after you actually change the typing constaint in the future, so the implementation can be adapted to the strategy at that time. For now I'll add a test for the current_time as you command :)

neutrinoceros:
>  Shall I set the default value as self.quan(-1, "s") right now? While I prefer to do it after you actually change the typing constaint in the future, so the implementation can be adapted to the strategy at that time.

So, the addition of type annotations in yt is a very long process that started a year ago and is far from complete.
The way I see it, yt_idefix is, to an extent, a miniature version of yt in terms of development. I'm trying to experiment on this sort of stuff here to see if it's viable or already too difficult to be ported to the mother ship, so I want yt_idefix to be at leas as constrained as yt itself, but even more so where possible.

In short, yes, I'd like to do this change now. I do not intend to make any change upstream that would break this strategy.

xshaokun:
I respect your plan and I need more time to think about how to make it. Because it means we give up `self._set_derived_attrs()` assigning proper code_time to `self.current_time`, we have to change its unit to code_time at a right place.

tests/data/pluto_disk_planet/meta.yaml
```diff
@@ -5,6 +5,7 @@ attrs:
   geometry: spherical
   dimensionality: 3
   kind: pluto
+  def_units: True
```
neutrinoceros, line 8: Seems like a good solution, but I'm not sure the variable name is clear enough. If you mean to indicate that units can be parsed from `definitions.h`, please rename this key to `"has_units"` to follow the same naming convention as `"has_geometry"`.
Also it'd be good to add this parameter to idefix datasets `meta.yaml` files (you can set it to False every where there)

xshaokun:
Sorry, I should've report my progress sooner.

I have to say, setting the default value as self.quan(-1, "s") might be not a good choice, since our logging would therefore be something like 

    yt : [INFO     ] 2022-02-07 02:07:39,898 Parameters: current_time              = 1.0 s

which is only true when code_time = 1 s. **We should ensure the information in logging message is correct.** So I tried to find a solution to set default value with unit of code_time. That means I have to figure out why code_time is always 1 s as showed by my script (refer to the output in my first comment in this PR) and solve it.

And now I find out it's caused by `self.registry._unit_object_cache`. 

`self.quan(1, "code_unit")` gets the unit information from unit_registry at first time and store it in `self.registry._unit_object_cache`. So every time we call `Unit("code_time", registry)` after we initialized `self.current_time` with `self.quan(1, "code_time")`, we always get code_time from this unit cache ( refer to https://github.com/yt-project/unyt/blob/ec5b3301c110787c9a67c600a66d8051ab8c831e/unyt/unit_object.py#L222 and https://github.com/yt-project/unyt/blob/ec5b3301c110787c9a67c600a66d8051ab8c831e/unyt/array.py#L564).

I think, technically the unit cache should update after the unit in unit_registry is modified, but the developors didn't consider such scenario. So do you think it's a bug or a feature requst for unyt ? Considering you plan to settle the type of current_time, I think this feature is necessary.

xshaokun:
> technically the unit cache should update after the unit in unit_registry is modified

For now, I think I can manually update the cache. I'll implement it tomorrow.

neutrinoceros:
Just a rapid feedback: unyt isn't actively being developed or maintained at the moment, so whatever unfortunate behavior it may have, we should consider it unchangeable and need to work around it.
Also I didn't intend to require you to work 3 times as much on this PR just to satisfy my own standards (which I'm sure I violated myself many times in this and other projects), so I'm apologize for that. It's fine to keep a temporary int value if it saves you time, especially if existing frontends do this as well.

In any case it sounds like we don't understand exactly what's happening yet, so we probably need to dig a bit more here. Hopefully I'll have some time to give you more hints in a couple days.

xshaokun:
I might not explain clearly last night, but I truly fix the bug in this commit and intialized the current_time with code_time. Although I make it complicated last night, we only need to clear the unit cache :)

> Also I didn't intend to require you to work 3 times as much on this PR just to satisfy my own standards (which I'm sure I violated myself many times in this and other projects), so I'm apologize for that.

I think it's good. Especially you are the owner and the one who makes the most efforts to matain the code, writing the code in your own standards is helpful, ( if I think it makes sense haha). So no need to apologize.

I have really learned many things from you and this project, such as some tools to matain a project like pytest and code styling. Therefore, _please don't hesitate to give any coments or even criticize me if necessary_ :)

yt_idefix/data_structures.py
```diff
@@ -344,7 +344,7 @@ def _parse_parameter_file(self):
             [fdata[f"xr{idir}"][-1] for idir in "123"], dtype="float64"
         )
 
-        self.current_time = fdata["time"]
+        self.current_time = self.quan(fdata["time"], "code_time")
```
neutrinoceros, line 347: Just to be sure: this is in the `IdefixDmpDataset` class, which isn't a parent to `PlutoVtkDataset`, so it's not directly related to the bug you were addressing, but it's a similar issue that you're fixing while you're at it, correct ?

neutrinoceros (commented):
I've checked that your tests are correctly pointing out the issue (they fail without the associated patch).
Now I don't really understand why the patch works (disclaimer: haven't had time to dive into it yet) but it seems like it's actually a fix that could maybe be upstreamed to yt ? if you still think that it's really a unyt bug, please report it there and link the issue in a comment next to the change so we know when we'll be able to clean it up (if ever).
Otherwise I'll just review this again much more carefully in a couple days and see if I agree with your approach.

yt_idefix/data_structures.py
xshaokun: Oh I did it because you said you want current_time to be unyt_quantity intially, so I thought you forgot to make it here. Since idefix doesn't include unit definitions, the bug I'm addressing is not a bug indeed for idefix dataset.

neutrinoceros: it's fine, just wanted to make sure this wasn't intended for `IdefixVtkDataset`. Che che !

xshaokun:
> Now I don't really understand why the patch works

In order to speeding up unit conversion, the units were stored in `self.unit_registry._unit_object_cached` once they were called and would be obtained from this cache next time. I initialized `self.current_time = self.quan(-1, "code_time")` in `self._parse_parameter_file()` which is called right after the unit_registry was created. So the code units were all with default values ( see [self._create_unit_registry()](https://github.com/yt-project/yt/blob/b35942193e28bfd61e606e575118b0a864062f33/yt/data_objects/static_output.py#L1162)`) and the default code_time was stored in the cache since then. 

The bug is because every time we picked the unit which is from the cache instead of from the registry, so it's always with default value.Therefore my solution is to clear the cache, then the unit will be picked from the registry ( with updated value) and stored in cache again. 

Not sure if I have explained it clearly. About the mechanism how to get a unit, you can refer to the definition `unyt.Unit()` when you have some time, and `self.quan()` gets the unit through `Unit()` as well.

> but it seems like it's actually a fix that could maybe be upstreamed to yt ? 

It's exactly what I meant in my previous comment right before my last commit ! It can be migrated to upstream if you think it's necessary. While I think we might be the only users who called the code units before they are set up properly.

neutrinoceros:
> It's exactly what I meant in my previous comment right before my last commit !

I'm sorry, I'm a bit overwhelmed with context switching right now. Ok I'll dedicate a moment to properly learning what's happening here sometimes this week.

yt_idefix/data_structures.py
xshaokun: Btw, I'm curious about what does `che che` mean? Is that a French phrase or english? 😆 Just cannot find its meanning.

neutrinoceros: Ha ! That was a lazy attempt at using the latin alphabet for "谢谢" ;)

neutrinoceros (approved):
Ok I finally looked carefully into this problem and I saw that the base `Dataset` class is designed around the idea that we can use a float as a place holder value in `self.current_time` (during `_parse_parameter_file`) so that its conversion to a `unyt_quantity` instance is delayed until after `code_time` has been properly registered. I realise now that adding a strict type constraint on this argument is a bad idea for now, I'm really sorry about that, I should have been more careful when I first commented.

I'll make the appropriate changes on your branch and squash-merge it. Thank you very much for bringing my attention here and for coming up with a solution even under my ill-informed advices !

PR #93, opened 2022-02-09
ENH: enable to set default_species_fields

xshaokun:
The setup of nuclei density fields can be activated by setting `default_species_fields="ionized"`. The derived fields `H_nuclei_density` and `El_number_density` are useful, for example, in emission plotting.

So I intended to add a keyword parameter `default_species_fields` in our frontend.

### DIFF
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -120,6 +120,7 @@ class IdefixDataset(Dataset, ABC):
         *,
         geometry: str | None = None,
         inifile=None,
+        default_species_fields=None,
     ):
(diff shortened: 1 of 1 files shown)


neutrinoceros (approved):
Cool ! This also made me realise that currently, IdefixDataset.__init__ (and its children classes) have incompatible signatures with the parent's class `Dataset`
```python
    def __init__(
        self,
        filename,
        dataset_type=None,
        file_style=None,
        units_override=None,
        unit_system="cgs",
        default_species_fields=None,
    ):
```
I'll accept your PR as is and make a follow up change to fix that. Thank you !

PR #94, opened 2022-02-09
BUG: fix IdefixDataset.__init__ signature for compatibility with parent class

neutrinoceros:
requesting a quick review from @xshaokun 
Can you please tell me if there's any other valid value to the `default_species_fields` argument besides `"ionized"` and `None` ?


xshaokun:
It can also be passed "neutral". I think that's all. Because I see other value would raise error.

neutrinoceros:
Excellent, thanks ! I'm going to experiment with typing here before I try to upstream more annotations to yt.

PR #120, opened 2022-07-02
EXP: plugin new StretchedGrid class

neutrinoceros:


neutrinoceros:
@matthewturk I'm experimenting with your StrechedGrid PR here. We'll probably be able to troubleshoot many small issues with it !

Here's a simple reprod to expose  a unit problem
```python
import yt
import yt_idefix

ds = yt.load("tests/data/big_mhdisc/dump.0015.dmp")
yt.SlicePlot(ds, "phi", ("gas", "density"))
```

neutrinoceros:
Current blocker: projections are still incorrect.

neutrinoceros:
aaaaand now they work !

PR #153, opened 2022-10-07
MNT: drop support for yt 4.0.x

neutrinoceros:


neutrinoceros:
I'll hold off for a couple weeks until I'm confident that dropping yt 4.0.x doesn't harm users

neutrinoceros:
I'm seeing some regressions in yt 4.1.0, so I'm going to hold off a little longer

PR #154, opened 2022-10-08
TST: require unyt>=2.9.2 in tests

neutrinoceros:


neutrinoceros:
type check errors are unrelated and fixed in #153

PR #167, opened 2023-01-06
ENH: resolve the variables in unit definitions (PLUTO)

### LINKED ISSUE #83: TODO: parse units and input parameters in unit definitions (Pluto)

xshaokun:
resolve #83

### DIFF
--- a/tests/data/pluto_disk_planet/definitions.h
+++ b/tests/data/pluto_disk_planet/definitions.h
@@ -28,11 +28,8 @@
 /* [Beg] user-defined constants (do not change this line) */
 
 #define  LIMITER                        VANLEER_LIM
-#define  UNIT_LENGTH                    7.779090384e13
-// #define  UNIT_LENGTH                    (5.2*CONST_au)
-#define  UNIT_DENSITY                   4.24857748e-9
-// #define  UNIT_DENSITY                   (CONST_Msun/(UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH))
-#define  UNIT_VELOCITY                  208457.85579
-// #define  UNIT_VELOCITY                  (sqrt(CONST_G*g_inputParam[Mstar]*CONST_Msun/UNIT_LENGTH)/(2.*CONST_PI))
+#define  UNIT_LENGTH                    (5.2*CONST_au)
+#define  UNIT_DENSITY                   (CONST_Msun/(UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH))
+#define  UNIT_VELOCITY                  (sqrt(CONST_G*g_inputParam[Mstar]*CONST_Msun/UNIT_LENGTH)/(2.*CONST_PI))
 
 /* [End] user-defined constants (do not change this line) */
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -576,8 +576,25 @@ class PlutoVtkDataset(IdefixVtkDataset):
             elif (unit_match := re.fullmatch(unit_regexp, line)) is not None:
                 unit = unit_match.group(1).lower() + "_unit"
                 expr = unit_match.group(2)
+                # Before evaluating the expression, replace the input parameters,
+                # pre-defined constants, code units and sqrt function
+                # that cannot be resolved. The order doesn't matter.
+                expr = re.sub(r"g_inputParam\[(\w+)\]", self._get_input_parameter, expr)
+                expr = re.sub(r"CONST_\w+", self._get_constants, expr)
+                expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
+                expr = re.sub(r"sqrt", "np.sqrt", expr)
                 expr = re.sub(constexpr, self._get_constants, expr)
                 self.parameters[unit] = eval(expr)
+
+    def _get_input_parameter(self, match: re.Match) -> str:
+        """Replace matched input parameters with its value"""
+        key = match.group(1)
+        return str(self.parameters["Parameters"][key])
+
+    def _get_unit(self, match: re.Match) -> str:
+        """Replace matched unit with its value"""
+        key = match.group(1).lower() + "_unit"
+        return str(self.parameters.get(key, 1.0))
 
     def _get_constants(self, match: re.Match) -> str:
         """Replace matched constant string with its value"""


neutrinoceros:
Thank you ! the pre-commit failure are unrelated, and I'm making it stable again in #168
I'll review this shortly

xshaokun:
Just rebase to the latest commit :)

yt_idefix/data_structures.py
```diff
@@ -519,6 +531,19 @@ def _get_constants(self, match: re.Match) -> str:
         key = match.group()
         return str(pluto_def_constants[key])
 
+    def _get_input_parameter(self, match: re.Match) -> str:
+        """Replace matched input parameter with its value"""
+        key = match.group(1)
+        if self.inifile is None:
+            self.inifile = os.path.join(self.directory, "pluto.ini")
```
neutrinoceros, line 538: This seems fragile for the following reasons:
1) Pluto supports giving custom names for this file (`./pluto -i myfile.ini` is valid) so we can't in general assume the name of this file. Even if it is present, it might not be the one that was actually used in the simulation results we are parsing.
2) we should be able to impose when the inifile is parsed, and alleviate some of the complexity here.

because of 1) (also true for Idefix btw), I don't think we should _ever_ assume the name of the inifile, and only parse it if it was explicitly set.

yt_idefix/data_structures.py
```diff
@@ -506,7 +510,15 @@ def _parse_header_file(self):
...
-                    expr = re.sub(constexpr, self._get_constants, expr)
+                    # Before evaluate the expression, replace the input parameters,
+                    # pre-defined constants, code units and sqrt function
+                    # that cannot be resolved.
+                    expr = re.sub(
+                        r"g_inputParam\[(\w+)\]", self._get_input_parameter, expr
+                    )
+                    expr = re.sub(r"CONST_\w+", self._get_constants, expr)
+                    expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
+                    expr = re.sub(r"sqrt", "np.sqrt", expr)
```
neutrinoceros, line 521: Does order matter ? It doesn't seem to, but in any case please add a comment to make it explicit.

yt_idefix/data_structures.py
```diff
@@ -280,6 +281,10 @@ def _parse_inifile(self) -> None:
         if self.inifile is None:
             return
 
+        if "Grid" in self.parameters:
+            # In case inifile was already parsed
+            return
+
```
neutrinoceros, line 287: If we end up parsing multiple times then the problem is elsewhere, I don't think we should guard against it here.

neutrinoceros (changes requested):
Seems good overall, but I'd like to refine the implementation details to avoid piling up complexity.

yt_idefix/data_structures.py
xshaokun: I think some default values would make it more convenient and follow the same tradition as Pluto.

Just like when users don't use a custom name for inifile, the additional option `-i myfile.ini` is not necessary and the command `./pluto` still works. 

So far this part is called _only when_ the unit definitions require the inifile. If the users use `pluto.ini` as the inifile and don't need to set it explicitly when loading the data, it still works as Pluto. Maybe it's more friendly for the Pluto users.

yt_idefix/data_structures.py
xshaokun: The comment I left here might be not exact. 

Since I call `self._parse_inifile()` in advance (but not always) in `self._parse_header_file()` which is called in `self.__init__()`, after which, however, the original `self._parse_inifile()` is called. Therefore here I avoid the inifile is parsed again.

If we could just only parse the inifile in `self.__init__()`, the problem will not exist.

yt_idefix/data_structures.py
xshaokun: Okay, I'll improve it.

yt_idefix/data_structures.py
neutrinoceros: I agree it's more friendly _when it works_, I'm arguing that it's more confusing _when it doesn't_. Maybe a compromise is to parse `pluto.ini` implicitly if and only if no other `*.ini` are found.

yt_idefix/data_structures.py
neutrinoceros: Then let's do that

yt_idefix/data_structures.py
xshaokun: I respect your opinion. And if we decided to parse the inifile in `self.__init__()` since then, Idefix users should also be considered. You can check my next commit :)

yt_idefix/data_structures.py
xshaokun: I found the best place to parse the inifile once is after `self.parameters` being created and before definitions.h being read, which should be somewhere in `Dataset.__init__()`. So I think we can't avoid this part.

neutrinoceros:
@xshaokun I realise that implementing the details of adequately discovering both inifiles and definitions headers and parsing them exactly once is a larger task than this one, and because I would like to get rid of it, I'm going to do it myself in a different PR that will take priority over yours. Would you agree to review it ?

xshaokun:
@neutrinoceros No problem!

xshaokun:
Thank @neutrinoceros for your work! So this PR only needs to make a tiny implementation.

yt_idefix/data_structures.py
```diff
@@ -576,9 +576,26 @@ def _parse_definitions_header(self) -> None:
             elif (unit_match := re.fullmatch(unit_regexp, line)) is not None:
                 unit = unit_match.group(1).lower() + "_unit"
                 expr = unit_match.group(2)
+                # Before evaluating the expression, replace the input parameters,
+                # pre-defined constants, code units and sqrt function
+                # that cannot be resolved. The order doesn't matter.
+                expr = re.sub(r"g_inputParam\[(\w+)\]", self._get_input_parameter, expr)
+                expr = re.sub(r"CONST_\w+", self._get_constants, expr)
+                expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
```
neutrinoceros, line 584: I *think* the group serves no purpose here, does it ?
```suggestion
                expr = re.sub(r"UNIT_\w+", self._get_unit, expr)
```

yt_idefix/data_structures.py
```diff
@@ -576,9 +576,26 @@ def _parse_definitions_header(self) -> None:
             elif (unit_match := re.fullmatch(unit_regexp, line)) is not None:
                 unit = unit_match.group(1).lower() + "_unit"
                 expr = unit_match.group(2)
+                # Before evaluating the expression, replace the input parameters,
+                # pre-defined constants, code units and sqrt function
+                # that cannot be resolved. The order doesn't matter.
+                expr = re.sub(r"g_inputParam\[(\w+)\]", self._get_input_parameter, expr)
```
neutrinoceros, line 582: Same as below
```suggestion
                expr = re.sub(r"g_inputParam\[\w+\]", self._get_input_parameter, expr)
```

yt_idefix/data_structures.py
```diff
@@ -576,9 +576,26 @@ def _parse_definitions_header(self) -> None:
...
+                expr = re.sub(r"CONST_\w+", self._get_constants, expr)
+                expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
+                expr = re.sub(r"sqrt", "np.sqrt", expr)
                 expr = re.sub(constexpr, self._get_constants, expr)
                 self.parameters[unit] = eval(expr)
 
+    def _get_input_parameter(self, match: re.Match) -> str:
+        """Replace matched input parameters with its value"""
+        key = match.group(1)
+        return str(self.parameters["Parameters"][key])
```
neutrinoceros, line 592: I don't understand where `self.parameters["Parameters"]` gets defined. Could you clarify ?

tests/data/pluto_disk_planet/definitions.h
```diff
@@ -28,11 +28,8 @@
...
 #define  LIMITER                        VANLEER_LIM
-#define  UNIT_LENGTH                    7.779090384e13
-// #define  UNIT_LENGTH                    (5.2*CONST_au)
-#define  UNIT_DENSITY                   4.24857748e-9
-// #define  UNIT_DENSITY                   (CONST_Msun/(UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH))
-#define  UNIT_VELOCITY                  208457.85579
-// #define  UNIT_VELOCITY                  (sqrt(CONST_G*g_inputParam[Mstar]*CONST_Msun/UNIT_LENGTH)/(2.*CONST_PI))
+#define  UNIT_LENGTH                    (5.2*CONST_au)
+#define  UNIT_DENSITY                   (CONST_Msun/(UNIT_LENGTH*UNIT_LENGTH*UNIT_LENGTH))
+#define  UNIT_VELOCITY                  (sqrt(CONST_G*g_inputParam[Mstar]*CONST_Msun/UNIT_LENGTH)/(2.*CONST_PI))
```
neutrinoceros, line 33: are we testing that a the result of parsing is unchanged ?

yt_idefix/data_structures.py
xshaokun: I need the group in `self._get_constants()`, instead of the whole match, same as above.

yt_idefix/data_structures.py
xshaokun: They're the input parameters given in the inifile, so they're parsed in `PlutoVtkDataset._parse_inifile()`. Should I add a comment here?

tests/data/pluto_disk_planet/definitions.h
xshaokun: Yes. Due to such definitions cannot be resolved before, I left their numerical results which are given in `tests/data/pluto_disk_planet/meta.yaml`

yt_idefix/data_structures.py
neutrinoceros: fair enough !

tests/data/pluto_disk_planet/definitions.h
neutrinoceros: excellent, thanks !

yt_idefix/data_structures.py
neutrinoceros: oh, of course they are. No need for adding a comment, thank you !

neutrinoceros (approved):

PR #169, opened 2023-01-07
ENH: automatic discovery and parsing for pluto.ini, idefix.ini, definitions.h and definitions.hpp

neutrinoceros:
This is in preparation for #167

### DIFF
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -7,6 +7,7 @@ import warnings
 import weakref
 from abc import ABC, abstractmethod
 from functools import cached_property
+from pathlib import Path
 from typing import Literal
 
 import inifix
@@ -205,6 +206,8 @@ class IdefixDataset(Dataset, ABC):
 
     _version_regexp = re.compile(r"v\d+\.\d+\.?\d*[-\w+]*")
     _dataset_type: str  # defined in subclasses
+    _default_definitions_header = "definitions.hpp"
+    _default_inifile = "idefix.ini"
 
     def __init__(
         self,
@@ -218,11 +221,18 @@ class IdefixDataset(Dataset, ABC):
         geometry: Literal["cartesian", "spherical", "cylindrical", "polar"]
         | None = None,
         inifile: str | os.PathLike[str] | None = None,
+        definitions_header: str | os.PathLike[str] | None = None,
     ):
         self._geometry_from_user = geometry
 
         dt = type(self)._dataset_type
         self.fluid_types += (dt,)
+
+        self._input_filename: str = os.fspath(filename)
+        self._inifile = self._get_meta_file(inifile, default=self._default_inifile)
+        self._definitions_header = self._get_meta_file(
+            definitions_header, default=self._default_definitions_header
+        )
 
         super().__init__(
             filename,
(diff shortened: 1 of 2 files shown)


neutrinoceros:
@xshaokun feel free to comment this. I still have a couple embedded TODOs to go over and a couple warnings to write, but I think the structure is ready.

yt_idefix/data_structures.py
```diff
@@ -232,14 +252,58 @@ def __init__(
...
+        elif self._header_type is Header.CPP:
+            ext = "hpp"
+        else:
+            assert_never(self._header_type)
+
+        root_dir = Path(self.directory)
+        if (
+            len(candidates := list(root_dir.glob(f"*.{ext}"))) == 1
+            and (file := candidates[0]).name == f"definitions.{ext}"
+        ):
```
xshaokun, line 300: As I know, the header file `definitions.h` cannot be replaced with another file by users in Pluto unless modifying the source code, don't know about the idefix. Is it enough to just check the existence of `definitions.h` or `definitions.hpp` here?

yt_idefix/data_structures.py
```diff
@@ -443,76 +530,36 @@ class PlutoVtkDataset(IdefixVtkDataset):
...
-                if unit_match is not None:
-                    unit = unit_match.group(1).lower() + "_unit"
-                    expr = unit_match.group(2)
-                    expr = re.sub(constexpr, self._get_constants, expr)
-                    self.parameters[unit] = eval(expr)
-        else:
-            warnings.warn(
-                f"Header file {self._definitions_header} couldn't be found. "
-                "The code units are set to be 1.0 in cgs by default."
-            )
```
xshaokun, line 515: I think this warning is better to be reserved to remind users that they maybe forget the header file since it's common to define the code units in Pluto.

yt_idefix/data_structures.py
neutrinoceros: Ah, thanks for pointing this out. Idefix indeed allows this file to be named arbitrarily, so I assumed Pluto did too. Do you think it would impact the user experience if I keep it that way still, so as to reduce code duplication ?

yt_idefix/data_structures.py
neutrinoceros: I'm removing it because with my patch, defining a non-existent header file will result in a FileNotFoundError on `open`, so the warning is unreachable. Does it make sense ?

xshaokun: How about showing this warning after checking the `self._definitions_header` is None, and then return?

yt_idefix/data_structures.py
xshaokun: I think it's fine!

yt_idefix/data_structures.py
neutrinoceros: does https://github.com/neutrinoceros/yt_idefix/pull/169/commits/b3857295f8019bbfd11177fcae5fc662ad0064e5 work for you ?

xshaokun: Yes, that's great!

neutrinoceros:
@xshaokun thank you for your feedback. I'll will now merge this so you can rebase your PR whenever suits you !

PR #174, opened 2023-01-08
ENH: parse geometry from definitions.hpp (idefix)

### LINKED ISSUE #171: ENH: implement _parse_definitions_header (Idefix)
wait for #169

neutrinoceros:
fix #171

follow up to #169 and #167


neutrinoceros:
For information, ping @xshaokun : I am slightly changing the code from your latest PR so that `_parse_definitions_header` doesn't create key/values pairs directly in `self.parameters` but in a sub dictionary `self.parameters["definitions"]` instead.
I plan on doing a 1.0.0 release with this patch tomorrow unless I think of something else that should be done in the mean time, in which case I'll release under 0.15.0

yt_idefix/data_structures.py
```diff
@@ -583,8 +593,8 @@ def _parse_definitions_header(self) -> None:
                 expr = re.sub(r"CONST_\w+", self._get_constants, expr)
                 expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
                 expr = re.sub(r"sqrt", "np.sqrt", expr)
-                expr = re.sub(constexpr, self._get_constants, expr)
-                self.parameters[unit] = eval(expr)
+                expr = re.sub(_CONST_REGEXP, self._get_constants, expr)
```
xshaokun, line 596: Ah sorry, I just noticed that I forgot to remove this line in my previous PR. It's the same as line#593 and now I prefer to make it explicit here.

yt_idefix/data_structures.py
```diff
@@ -40,6 +40,9 @@
     ZCoords = np.ndarray
 
 ytLogger = logging.getLogger("yt")
+_DEF_GEOMETRY_REGEXP: Final = re.compile(r"^\s*#define\s+GEOMETRY\s+([A-Z]+)")
+_DEF_UNIT_REGEXP: Final = re.compile(r"^\s*#define\s+UNIT_(\w+)\s+(\S+)")
+_CONST_REGEXP: Final = re.compile(r"CONST_\w+")
```
xshaokun, line 45: I think `_CONST_REGEXP` is no more necessary :)

xshaokun (changes requested):
Got it, thanks! I found a mistake I made in my last PR, I think you can help me fix it in this PR.

yt_idefix/data_structures.py
neutrinoceros: good catch, thanks !

yt_idefix/data_structures.py
```diff
@@ -315,26 +317,38 @@ def _parse_parameter_file(self):
...
             raise ValueError(
-                "Geometry couldn't be parsed from file. "
+                "Geometry couldn't be parsed from disk. "
                 "The 'geometry' keyword argument must be specified."
             )
-        elif from_user is not None:
-            if from_file is not None and from_user != from_user:
+
+        if from_input is not None:
+            if len({from_disk, from_input} - {None}) > 1:
```
xshaokun, line 340:
```suggestion
            if from_disk and from_input:
```
Is it enough and more straightforward here?

yt_idefix/data_structures.py
```diff
@@ -353,13 +367,19 @@ def _parse_inifile(self) -> None:
             if any(_ != "u" for _ in vals[3::3]):
                 msg_elems.append(f"found non-uniform block(s) in direction {ax}")
 
-    @abstractmethod
     def _parse_definitions_header(self) -> None:
+        self.parameters["definitions"] = {}
```
xshaokun, line 371: I see you defined `self.parameters["definitions"]` both here and in `PlutoVtkDataset`. Is it plausible to define it once in their super class `IdefixDataset`, before calling `_parse_definitions_header`

xshaokun (commented):
Just a few comments here :)

yt_idefix/data_structures.py
neutrinoceros: I don't think so. The more explicit way to implement my check is
```python
if from_disk is not None and from_input is not None and from_disk != from_input:
```
maybe I should use that instead if you think the current state is confusing. This is a case where having variable that can be more than one type (`str` or `None`) quickly escalates into nonesense... maybe I could just use `""` as default value instead of None 🤔

yt_idefix/data_structures.py
neutrinoceros: Well it's a tradeoff. I think it's probably better to repeat this one line rather than coupling the classes even more and make the flow ever so slightly harder to follow (admittedly it's currently a mess).

xshaokun: Ok!

xshaokun (approved):

neutrinoceros:
thank you @xshaokun for your help, I really appreciate it !

PR #187, opened 2023-01-24
BUG: add support for boolean fields in dump files

neutrinoceros:
TODO:
- [x] upstream this in idefix itself (currently only supported in my personnal fork)
- [x] make sure vtk files are supported too


neutrinoceros:
rebranded as a bugfix after I realised this is _already_ supported in mainline idefix

PR #202, opened 2023-02-20
Added PLUTO xdmf dump support

dutta-alankar:
Support for PLUTO xdmf (hdf5+xmf) output. Currently not tested extensively.

### DIFF
--- a/yt_idefix/_io/commons.py
+++ b/yt_idefix/_io/commons.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from enum import IntEnum
 from typing import Any, Dict, Literal, NamedTuple, Tuple
 
 import numpy as np
@@ -55,13 +54,3 @@ IdefixFieldProperties = Dict[str, Tuple[Prec, Dim, np.ndarray]]
 
 # Map various str keys to scalars and arrays
 IdefixMetadata = Dict[str, Any]
-
-
-class ByteSize(IntEnum):
-    """Constant byte size for data types"""
-
-    CHAR = 1
-    INT = 4
-    FLOAT = 4
-    DOUBLE = 8
-    BOOLEAN = 1
--- a/yt_idefix/_io/dmp_io.py
+++ b/yt_idefix/_io/dmp_io.py
@@ -7,7 +7,7 @@ from typing import BinaryIO, Literal, cast, overload
 
 import numpy as np
 
-from .commons import ByteSize, Dim, IdefixFieldProperties, IdefixMetadata, Prec
+from .commons import Dim, IdefixFieldProperties, IdefixMetadata, Prec
 
 KNOWN_GEOMETRIES: dict[int, str] = {
     1: "cartesian",
@@ -27,17 +27,11 @@ class CharCount(IntEnum):
 # enum DataType {DoubleType, SingleType, IntegerType};
 DTYPES: dict[int, Prec] = {0: "d", 1: "f", 2: "i", 3: "?"}
 DTYPES_2_NUMPY: dict[Prec, str] = {"d": "=f8", "f": "=f4", "i": "=i4"}
-DTYPES_2_SIZE: dict[Prec, ByteSize] = {
-    "i": ByteSize.INT,
-    "f": ByteSize.FLOAT,
-    "d": ByteSize.DOUBLE,
-    "?": ByteSize.BOOLEAN,
-}
 
 
 def read_null_terminated_string(fh: BinaryIO, maxsize: int = CharCount.NAME) -> str:
-    """Read maxsize * ByteSize.CHAR bytes, but only parse non-null characters."""
-    b = fh.read(maxsize * ByteSize.CHAR)
+    """Read maxsize bytes, but only parse non-null characters."""
+    b = fh.read(maxsize)
     s = b.decode("utf-8", errors="backslashreplace")
     s = s.split("\x00", maxsplit=1)[0]
     return s
@@ -130,7 +124,7 @@ def read_chunk(
     # type hints for now
     assert ndim == len(dim)
     count = np.product(dim)
-    size = count * DTYPES_2_SIZE[dtype]
+    size = count * np.dtype(dtype).itemsize
     if skip_data:
         fh.seek(size, 1)
         return None
@@ -224,7 +218,7 @@ def get_field_offset_index(fh: BinaryIO) -> dict[str, int]:
     field_index = {}
 
     # skip header
-    fh.seek(CharCount.HEADER * ByteSize.CHAR)
+    fh.seek(CharCount.HEADER)
     # skip grid properties
     for _ in range(9):
         _field_name, dtype, ndim, dim = read_next_field_properties(fh)
@@ -264,7 +258,7 @@ def read_idefix_dump_from_buffer(
     fh: BinaryIO, skip_data: bool = False
 ) -> tuple[IdefixFieldProperties, IdefixMetadata]:
     # skip header
-    fh.seek(CharCount.HEADER * ByteSize.CHAR)
+    fh.seek(CharCount.HEADER)
 
     data: float | np.ndarray | None
     fprops: IdefixFieldProperties = {}
--- a/yt_idefix/_io/vtk_io.py
+++ b/yt_idefix/_io/vtk_io.py
@@ -6,7 +6,7 @@ from typing import Any, BinaryIO, Literal, overload
 
 import numpy as np
 
-from .commons import ByteSize, Coordinates, Shape
+from .commons import Coordinates, Shape
 
 KNOWN_GEOMETRIES: dict[int, str] = {
     0: "cartesian",
@@ -54,7 +54,7 @@ def read_single_field(
     if offset is not None and fh.tell() != offset:
         fh.seek(offset)
     if skip_data:
-        fh.seek(count * ByteSize.FLOAT, 1)
+        fh.seek(count * np.dtype("f").itemsize, 1)
         data = None
     else:
         data = np.fromfile(fh, ">f", count=count)


dutta-alankar:
I fixed the bugs that were there in the previous commit. I'm also attaching a [link to the data](https://drive.google.com/file/d/1yZlApXBKMI14bTNDF8IkgX3nh4Cy4FRx/view?usp=sharing) that I used for plotting. In the zip file test.py has the plotting script.

neutrinoceros:
Thanks for cleaning up your code.

Here's what needs to happen for this to be mergeable:
1) pass existing tests
2) conform to the style guide (this is validated by pre-commit.ci and can mostly be automated, see https://yt-project.org/docs/dev/developing/developing.html#coding-style-guide)
3) type check

You can set 3) aside for now, but 1) is crucial and 2) is should be easy enough, so let's get them out of the way !

Anticipating a bit but when everything is green, we'll also want to add a least one new test case. This will require committing a (preferably small) dataset to your branch, somewhere in `tests/data`, but it might also require some tweaks to the infrastructure. Anyway, let's not get ahead of ourselves.

Right now, tests are failing with
```
ImportError: cannot import name 'Geometry' from 'yt.geometry.api'
```

This is because the `Geometry` enum isn't included in yt `4.1.4`, which is the main version yt_idefix is currently tested against. However you shouldn't *need* to care about that, as this complexity is already handled in `IdefixDataset._setup_geometry`, so you just need to inherit this method.

I suggest verifying your code locally before pushing. For lack of a contribution guide, I'll give you a couple pointers for how to do that here (I'm assuming all commands are run from the project directory)

1) use a virtualenv and install yt_idefix there. In brief:
```shell
python -m venv venv
source venv/bin/activate
python -m pip install --editable .
```
2) Install test-dependencies as
```shell
python -m pip install -r requirements/tests.txt
```
3) run tests locally
```shell
pytest
```

If you want to learn more about virtual environments, I suggest this excellent introduction https://realpython.com/python-virtual-environments-a-primer/

dutta-alankar:
Thanks a lot. I'll get started with your suggestions rightaway.

dutta-alankar:
Using pytest on a new virtual environment, I get the following error message:
```
_
tests/test_C_io.py:3: in <module>
    from yt_idefix._io import C_io
yt_idefix/__init__.py:3: in <module>
    from yt_idefix.api import *
yt_idefix/api.py:2: in <module>
    from .data_structures import (
yt_idefix/data_structures.py:16: in <module>
    from yt.data_objects.index_subobjects.stretched_grid import StretchedGrid
../../yt/yt/__init__.py:12: in <module>
    import yt.utilities.physical_constants as physical_constants
../../yt/yt/utilities/physical_constants.py:3: in <module>
    from yt.units.yt_array import YTQuantity
../../yt/yt/units/yt_array.py:3: in <module>
    from yt.funcs import array_like_field  # NOQA: F401
../../yt/yt/funcs.py:32: in <module>
    from yt.utilities.on_demand_imports import _requests as requests
../../yt/yt/utilities/on_demand_imports.py:103: in <module>
    _netCDF4 = netCDF4_imports()
../../yt/yt/utilities/on_demand_imports.py:92: in __init__
    import netCDF4  # noqa F401
../../../.local/lib/python3.9/site-packages/netCDF4/__init__.py:3: in <module>
    from ._netCDF4 import *
src/netCDF4/_netCDF4.pyx:1: in init netCDF4._netCDF4
    ???
E   RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 16 from C header, got 96 from PyObject
```
What is the problem?

neutrinoceros:
I think this indicates that netCDF4 was compiled with a version of numpy that is newer than the one you have installed. Either uninstalling netCDF4 or upgrading numpy should fix it.

dutta-alankar:
> I think this indicates that netCDF4 was compiled with a version of numpy that is newer than the one you have installed. Either uninstalling netCDF4 or upgrading numpy should fix it.

Let me check this. In the meantime, now the new code seems to pass most of the tests.

neutrinoceros:
Good job ! I should be able to do a first round of review later today. Is it okay if I push one or two small patches to fix type checking, or would you prefer that I explain what's happening and how to solve it ?

dutta-alankar:
Following the data directory in your tests, I created similar xdmf outputs from PLUTO. It would be great if you can have a look at this and possibly help me in writing additional tests for these outputs. [Link is here](https://drive.google.com/file/d/1i9kqSq-YEzrtxeVRcrcROGu6wc83vcio/view?usp=sharing).

neutrinoceros:
Any chance you could produce a smaller dataset ? 330Mb is very heavy for git to track. I think the largest dataset that I'm currently storing in this repo is about 2Mb.
For reference, yt_idefix's test infrastructure is designed to automatically use any datasets that are organised as a subdirectory in `tests/data/` and contain a file named `meta.yaml`. This means you can test on large datasets locally, but I would like to avoid *commiting* large datasets if it can be helped.

dutta-alankar:
> Any chance you could produce a smaller dataset ? 330Mb is very heavy for git to track. I think the largest dataset that I'm currently storing in this repo is about 2Mb. For reference, yt_idefix's test infrastructure is designed to automatically use any datasets that are organised as a subdirectory in `tests/data/` and contain a file named `meta.yaml`. This means you can test on large datasets locally, but I would like to avoid _commiting_ large datasets if it can be helped.

Yes sure! All the data except the one in the polar directory is pretty small in size. So, one can skip that.

neutrinoceros:
I've looked at type-checking errors and my conclusion is that they are triggered because `PlutoXdmfDataset` inherits directly from `PlutoVtkDataset`, while it should inherit from a more generic `PlutoDataset` class that is currently missing; anyway, I can handle that myself later, let's put it aside for the time being. I'll do a first round of review now.

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+                    try:
+                        from yt.geometry.api import (
+                            Geometry,  # type: ignore [attr-defined]
+                        )
+
+                        return Geometry(geom)
+                    except ImportError:
+                        pass
+
+                return geom
```
neutrinoceros, line 878: this definition shouldn't be nested in the `with` block. Also it's duplicated from `IdefixDataset` and I don't think it's necessary.
Because this class inherits from `IdefixDataset` I would expect the geometry parsing to happen via the parent class, with `super()._parse_parameter_file()` (or `IdefixDataset._parse_parameter_file(self)` if you need to skip `IdefixVtkDataset`)
Please tell me if anything doesn't work with my proposed approach. The intention is to save a bunch of repetitions and simplify your own specialised class

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+
+        with open(out_file) as outttxt:
+            txt = outttxt.readlines()
+            entry = int(
+                os.path.basename(self.parameter_filename)
+                .replace(".flt.h5", "")
+                .replace("data.", "")
+                .replace(".dbl.h5", "")
+            )
+            self.current_time = float(txt[entry].split()[1])
```
neutrinoceros, line 891: again, can you explain how many fields are expected from that `split` class and what they are supposed to be ?

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+        self.mu = 0.61
+        self.storage_filename = self.parameter_filename
+        self.refine_by = 1  # no mesh refinement
+        self._periodicity = (True, True, True)
+        # PlutoStatic cannot yet be run as a cosmological simulation
+        self.cosmological_simulation = 0
+        self.current_redshift = 0.0
+        self.omega_lambda = 0.0
+        self.omega_matter = 0.0
+        self.hubble_constant = 0.0
```
neutrinoceros, line 909: You'd get this for free by calling the parent class as I mentioned in a previous comment

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+
+    def _set_code_unit_attributes(self):
```
neutrinoceros, line 911: Is there a reason to override PlutoVtkDataset's implementation of this method ?
They seem to accomplish the same thing, so I'd prefer to keep them identical unless they're an actual difference I am missing

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+            self.magnetic_unit = self.quan(magnetic_unit, "gauss")
+
+        for key, unit in self.__class__.default_units.items():
+            setdefaultattr(self, key, self.quan(1, unit))
+
+    @classmethod
+    def _is_valid(cls, filename, *args, **kwargs):
+        # This accepts a filename or a set of arguments and returns True or
+        # False depending on if the file is of the type requested.
+        test1 = ("dbl.h5" in filename) or ("flt.h5" in filename)
```
neutrinoceros, line 1018: In this method it's important to return as early as possible, so prefer this style
```suggestion
        if not ("dbl.h5" in filename) or ("flt.h5" in filename):
            return False
```

yt_idefix/fields.py
```diff
@@ -47,3 +50,247 @@ def setup_fluid_fields(self):
...
+    )
+
+    known_particle_fields = ()
+
+    def setup_fluid_fields(self):
+        unit_system = self.ds.unit_system
+
+        # Add tracer fields
+        for i in range(1, self.ds.ntracers + 1):
+            if (self.ds._dataset_type, "tr%d" % i) in self.field_list:
```
neutrinoceros, line 79: please avoid `%`-style formatted strings and prefer fstrings

yt_idefix/fields.py
```diff
@@ -47,3 +50,247 @@ def setup_fluid_fields(self):
...
+            self.alias(
+                ("gas", "number density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
+            self.alias(
+                ("gas", "Number density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
```
neutrinoceros, line 246: is there a precedent in yt for having so many aliases of the same field where only the case is varied ? Unless other frontends are doing it, I think it should be avoided.
By convention, spaces should also be replaced by underscores in field names.

yt_idefix/io.py
```diff
@@ -85,3 +87,42 @@ def _read_particle_coords(self, chunks, ptf):
...
         # idefix doesn't have particles (yet)
         raise NotImplementedError("Particles are not currently supported for Idefix")
+
+
+class PlutoXdmfIOHandler(BaseIOHandler):
+    _particle_reader = False
+    _dataset_type = "pluto-xdmf"
+
+    def _read_particle_coords(self, chunks, ptf):
+        raise NotImplementedError("Reading of particle fields not yet impleted!")
```
neutrinoceros, line 97:
```suggestion
        raise NotImplementedError("Reading of particle fields not yet implemented!")
```

yt_idefix/io.py
```diff
@@ -85,3 +87,42 @@ def _read_particle_coords(self, chunks, ptf):
...
+
+    def _read_particle_fields(self, chunks, ptf, selector):
+        raise NotImplementedError("Reading of particle fields not yet impleted!")
```
neutrinoceros, line 100:
```suggestion
        raise NotImplementedError("Reading of particle fields not yet implemented!")
```

yt_idefix/misc.py
```diff
@@ -0,0 +1,72 @@
+def remove_comments(input_text):
```
neutrinoceros, line 1: I think this all function becomes unneeded if we leverage inheritance as I suggested

neutrinoceros (commented):
Here's a first batch of comments and questions. Overall this seems like good work, but I think there's room to reduce the size of the patch by avoiding repetitions, and I'd like a couple magic numbers to be explained.

yt_idefix/data_structures.py
```diff
@@ -210,7 +217,7 @@ def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
         return cell_centers
 
 
-class IdefixDmpHierarchy(IdefixHierarchy):
+class IdefixDmpHierarchy(IdefixVtkHierarchy):
```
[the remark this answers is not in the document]
dutta-alankar: I think this was a typo that I accidentally changed.

yt_idefix/data_structures.py
```diff
@@ -236,6 +243,72 @@ def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
         )
 
 
+class PlutoXdmfHierarchy(IdefixHierarchy):
+    def _get_field_offset_index(self) -> dict[str, int]:
+        retv: dict[str, int] = {}
+        retv["dummy"] = 0
+        return retv  # Dummy
```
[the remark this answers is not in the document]
dutta-alankar: Yes exactly. I was wanting it to pass the type checking test.

yt_idefix/data_structures.py
```diff
@@ -236,6 +243,72 @@ def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
...
+
+    def _detect_output_fields(self):
+        with h5py.File(self.index_filename, mode="r") as h5f:
+            self.field_list = [
+                (self.ds._dataset_type, "%s" % k)
+                for k in h5f[list(h5f.keys())[0] + "/vars/"]
+            ]
```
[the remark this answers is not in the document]
dutta-alankar: I agree that f-formatted strings would be better and str(k) is better in case it is not string.

yt_idefix/data_structures.py
```diff
@@ -236,6 +243,72 @@ def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
...
+
+    def _parse_grid_data(self, gridtxt):
+        start = 10
+        nx1 = int(gridtxt[start][:-1])
+        start = start + (nx1 + 1)
```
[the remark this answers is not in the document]
dutta-alankar: Sure! Please change.

yt_idefix/data_structures.py
```diff
@@ -236,6 +243,72 @@ def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
...
+    def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
+        cell_center1, cell_center2, cell_center3 = self._cell_widths
+        start = cell_center1[0]
+        cell_center1 = start + 0.5 * cell_center1
+        start = cell_center2[0]
+        cell_center2 = start + 0.5 * cell_center2
+        start = cell_center3[0]
+        cell_center3 = start + 0.5 * cell_center3
+        cell_widths = (cell_center1, cell_center2, cell_center3)
+        return cell_widths
```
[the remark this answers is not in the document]
dutta-alankar: Yeah this change is better. I was just writing in this way as it helped to think of the logic at that time.

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+    _dataset_type = "pluto-xdmf"
+    _version_regexp = re.compile(r"\d+\.\d+\.?\d*[-\w+]*")
+    _required_header_keyword = "PLUTOXdmf"
+    _default_definitions_header = "definitions.h"
+    _default_inifile = "pluto.ini"
+    _field_offset_index = None
+
+    def _parse_parameter_file(self):
+        grid_file = os.path.join(os.path.dirname(self.parameter_filename), "grid.out")
+        self.parameter_filename[:-2] + "xmf"
```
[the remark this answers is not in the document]
dutta-alankar: I meant to keep a variable for the xmf file location in case it is used later. But I think since I'm currently not using it, I missed the assignment. Let this be assigned to a variable as xmf files have useful info that can be used later for something.

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+        with open(grid_file) as gridtxt:
+            txt = gridtxt.readlines()
+            self.domain_left_edge = np.array(
+                [
+                    float(
+                        txt[i]
+                        .replace(",", "")
+                        .replace("[", "")
+                        .replace("]", "")
+                        .split()[2]
```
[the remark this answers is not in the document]
dutta-alankar: I had to remove the , and the [ and ] for the split to work properly.

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+    _default_definitions_header = "definitions.h"
+    _default_inifile = "pluto.ini"
+    _field_offset_index = None
+
+    def _parse_parameter_file(self):
+        grid_file = os.path.join(os.path.dirname(self.parameter_filename), "grid.out")
+        self.parameter_filename[:-2] + "xmf"
+        out_file = os.path.join(
+            os.path.dirname(self.parameter_filename),
+            self.parameter_filename[-6:] + ".out",
```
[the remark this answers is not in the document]
dutta-alankar: This gets rid of few of the extensions at the end. So I thought directly slicing is good. However, let me think of something better.

yt_idefix/data_structures.py
```diff
@@ -724,3 +797,241 @@ def _validate_units_override_keys(cls, units_override):
...
+            self.domain_left_edge = np.array(
+                [
+                    float(
+                        txt[i]
+                        .replace(",", "")
+                        .replace("[", "")
+                        .replace("]", "")
+                        .split()[2]
+                    )
+                    for i in [6, 7, 8]
```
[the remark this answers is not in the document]
dutta-alankar: This is not optimal. I missed generalizing it for lower dimensions. These are the lines where grid.out has info about the grid.

yt_idefix/data_structures.py
dutta-alankar: There were other stuffs like reading vtk that was done by IdefixVtkDataset._parse_parameter_file(self). So I copied this part of the code. Maybe using IdefixDataset would fix this.

yt_idefix/data_structures.py
dutta-alankar: The filenames are data.%04d.dbl.h5 or data.%04d.flt.h5 
However, I didn't want the code to depend on %04d but to have something more general.

yt_idefix/data_structures.py
dutta-alankar: Checking node_coords and cell_coords ensure that it is not just any hdf5 file.

yt_idefix/fields.py
dutta-alankar: Yes I'll put f-strings.

yt_idefix/fields.py
dutta-alankar: I'm new to yt. So, please change this in however you deem appropriate.

yt_idefix/io.py
dutta-alankar: Fixed!

yt_idefix/misc.py
dutta-alankar: Yes sure! Let me see how to use the inheritance.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: Fixed!

yt_idefix/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: consider using `str.partition` to split all extensions at once in a more expressive and robust way
Here's a simple illustration of what it does
```python
>>> "a.b.c.d".partition(".")
('a', '.', 'b.c.d')
```

yt_idefix/data_structures.py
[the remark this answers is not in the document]
neutrinoceros: What I meant was, `split()` is returning a list of str, and you're taking only the third element, but I have no idea how many of them are expected, and what the others are expected to be. Ideally I would like the answers to my questions to be expressed in the code itself, or by a comment.

yt_idefix/fields.py
neutrinoceros: In that case I suggest sticking to one name per field (no aliases) and use conventional names from yt.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: ```grid.out``` is a file which has lines like the following from where I wanted to read extents and count.
# X1: [ 0.040000,  0.500000], 400 point(s), 2 ghosts
# X2: [ 0.000000,  6.283185], 400 point(s), 2 ghosts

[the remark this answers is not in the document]
dutta-alankar: I have made a few changes to have this more general and less hard coded. Please check and let me know if there are any changes that are needed.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: I have made edits so that it need not have this [6,7,8] array.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: The idea here is whenever there is a ```data.%04.dbl.h5``` file or a ```data.%04.flt.h5``` there are corresponding files ```dbl.h5.out``` and ```flt.h5.out``` respectively which needs to be read and parsed. Also there are companion files with names ```data.%04.dbl.xmf``` and ```data.%04.flt.xmf``` which are read by the code. Since these formats don't change the string slicings were hard coded. Can you suggest anything that's better?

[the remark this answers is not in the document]
neutrinoceros: > Since these formats don't change the string slicings were hard coded.

Fair point, but the reason I'm asking to avoid slicing is that it's hard to reason about for me as a mostly clueless maintainer, but also for you if you come back to this code in two years.
We could use `str.removesuffix('.h5.out')` here. This method requires Python 3.9 but I'm okay with dropping support for Python 3.8 now. Anyway this kind of technical details shouldn't clobber your PR. Can you just add a comment explaining *what 6 characters* you mean to remove, so I can refactor later with confidence ?

yt_idefix/data_structures.py
neutrinoceros: this warrants a comment then

yt_idefix/data_structures.py
neutrinoceros: I trust @xshaokun could answer that.

yt_idefix/data_structures.py
neutrinoceros: the extension is a necessary condition, not a sufficient one, but we can still return early if we get any other extension !
I'm not suggesting to throw away the rest of the function, just that we should performs checks as greedily as possible and return with the first invalidated necessary condition.

yt_idefix/data_structures.py
xshaokun: > I'm not sure if the logic of partial code units in definitions.h is already implemented.

This method in PlutoVtkDataset is able to handle the partial definitions of these three units. Namely, if any of them is not defined in definitions.h, the default value will be assigned as Pluto did. 

However I just realized the default values are wrong. Thanks for your remind! I will fix it in my next PR soon.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: > > Since these formats don't change the string slicings were hard coded.
> 
> Fair point, but the reason I'm asking to avoid slicing is that it's hard to reason about for me as a mostly clueless maintainer, but also for you if you come back to this code in two years. We could use `str.removesuffix('.h5.out')` here. This method requires Python 3.9 but I'm okay with dropping support for Python 3.8 now. Anyway this kind of technical details shouldn't clobber your PR. Can you just add a comment explaining _what 6 characters_ you mean to remove, so I can refactor later with confidence ?

Okay, I'll do that

yt_idefix/data_structures.py
```diff
@@ -724,3 +793,272 @@ def _validate_units_override_keys(cls, units_override):
...
+                    length_unit = float(length_unit)
+                except ValueError:
+                    for name in constant_names:
+                        length_unit = length_unit.replace(
+                            name, f'pluto_def_constants["{name}"]'
+                        )
+                    for name in inputParam_names:
+                        length_unit = length_unit.replace(name, f'"{name}"')
+                    length_unit = length_unit.replace("sqrt", "np.sqrt")
+                    length_unit = length_unit.replace("log", "np.log")
```
xshaokun, line 973: I'm not sure if `log` will be used in the unit definitions, could you show me an example? If it's possile, I will add it in my next PR too.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: These are the files needed and I have added comments accordingly>
# data.%04d.<dbl/flt>.h5 -> data.%04d.<dbl/flt>.xmf
# data.%04d.<dbl/flt>.h5 -> <dbl/flt>.h5.out

xshaokun (commented):
I think you can just use the original method after I fix one or two bugs :)

yt_idefix/data_structures.py
neutrinoceros: Feel free to open a separate PR for that.

yt_idefix/data_structures.py
dutta-alankar: I want to use only the ```parse_geometry``` function while not anything else in ```_setup_geometry```. Is there a way to acomplish this taking advantage of the inheritance and avoiding repetitions? This is because the rest of the code in ```_setup_geometry``` is not suited for the xdmf files and causes errors if called. @neutrinoceros  Can you suggest a solution?

dutta-alankar: I did a small change in the code in how parse geometry was placed. Let me know if this is okay.

yt_idefix/data_structures.py
dutta-alankar: Added!

yt_idefix/data_structures.py
dutta-alankar: ```IdefixDataset._parse_parameter_file(self)``` has other function calls like 
```
self._parse_inifile()
self._parse_definitions_header()
self._setup_geometry()
```
These functions are not there in xdmf reads and going to cause problems. Therefore, I avoided calling the inherited ```IdefixDataset._parse_parameter_file(self)```.

yt_idefix/data_structures.py
dutta-alankar: Yes this was the reason I had a separate code to parse the ```definitions.h```. Maybe my part of the code can be put in one of the parent classes to avoid code repetitions. @xshaokun and @dutta-alankar Please have a look around this region of code in ```PlutoXdmfDataset._set_code_unit_attributes(self)```.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: Also added a comment to be more clear.

yt_idefix/data_structures.py
xshaokun: As I see, your method is doing almost the same things as `PlutoVtkDataset._set_code_unit_attributs(self)`, except that unit definitions in `definitions.h` was parsed in `PlutoVtkDataset._parse_definitions_header(self)`

yt_idefix/data_structures.py
```diff
@@ -724,3 +793,277 @@ def _validate_units_override_keys(cls, units_override):
...
+                count, self.dimensionality
+            ):  # These are dummy, may need some fix
+                domain_left_edge[i] = 0.0
+                domain_right_edge[i] = 1.0
+                domain_dimensions[i] = 1
+            self.domain_left_edge = domain_left_edge
+            self.domain_right_edge = domain_right_edge
+            self.domain_dimensions = domain_dimensions
+
+            self.geometry = self._parse_geometry(geom_str)
```
xshaokun, line 901: I'm wondering why you have to explicitly assign `self.geometry` here, the parent class PlutoVtkDataset will take care of it.

yt_idefix/data_structures.py
```diff
@@ -318,6 +410,19 @@ def _parse_parameter_file(self):
         self._parse_definitions_header()
         self._setup_geometry()
 
+    def _parse_geometry(self, geom: str):
```
neutrinoceros, line 413: this could be a `@staticmethod`, since the `self` argument isn't used

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
             )
 
         super(cls, cls)._validate_units_override_keys(units_override)
+
+
+class PlutoXdmfDataset(PlutoVtkDataset):
+    _index_class = PlutoXdmfHierarchy
+    _field_info_class = PlutoXdmfFields
+    _dataset_type = "pluto-xdmf"
+    _version_regexp = re.compile(r"\d+\.\d+\.?\d*[-\w+]*")
```
neutrinoceros, line 825: This appears to be identical to the class attributed inherited from `PlutoVtkDataset`
```suggestion
```

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
...
+    _required_header_keyword = "PLUTOXdmf"
+    _default_definitions_header = "definitions.h"
```
neutrinoceros, line 827: same as above
```suggestion
```

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
...
+    _default_inifile = "pluto.ini"
```
neutrinoceros, line 828: same as above
```suggestion
```

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
...
+    _field_offset_index = None
```
neutrinoceros, line 829:
```suggestion
    _field_offset_index = {} # TODO(clm): clean this up
```

yt_idefix/data_structures.py
dutta-alankar: Okay! I'll change that.

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
...
+
+    def _parse_parameter_file(self):
```
neutrinoceros, line 831: two notes here:
- `self.parameter_filename` is a historical alias and should be replaced with `self.filename`
- a shortcut `self.directory` can be used to avoid repeating `os.path.dirname(self.parameter_filename)`

it would make this function easier to read to leverage these

yt_idefix/data_structures.py
```diff
@@ -724,3 +816,290 @@ def _validate_units_override_keys(cls, units_override):
...
+        grid.out file has entries like the following:
+
+            # DIMENSIONS: 2
+            # GEOMETRY:   POLAR
+            # X1: [ 0.040000,  0.500000], 400 point(s), 2 ghosts
+            # X2: [ 0.000000,  6.283185], 400 point(s), 2 ghosts
+
+        These lines need to be parsed to create the appropriate grid data structure in yt.
+        Splitting and extracting the numers become straightforward when '[', ']' and ',' characters are removed before the split.
+        """
```
neutrinoceros, line 866: please avoid using multiline strings as comments.
this should either be using `#` or be included in the docstring, whatever you like best

yt_idefix/data_structures.py
neutrinoceros: Agreed. But the parent class needs to be explicitly called

yt_idefix/data_structures.py
neutrinoceros: I suggest we move discussion to #204 and then come back here and simplify this one :)

yt_idefix/data_structures.py
dutta-alankar: Okay then where do you suggest putting in the rest of the check?

neutrinoceros: Let me clarify. What I'm suggesting is to change how the function is organised to adopt the following form:
```python
if not necessary_condition_1:
    return False
if not necessary_condition_2:
    return False
if not necessary_condition_3:
    return False
...
return True
```
or even simpler (and not any less optimal)
```python
if not (
    necessary_condition_1
    and necessary_condition_2
    and necessary_condition_3
):
    return False
...
return True
```
of course some checks may be more complex to express than others (in particular, opening the file with h5py), but that's the gist.

yt_idefix/data_structures.py
dutta-alankar: Yes, for example one of our code we were using M_star as a parameter and using the M_star-M_halo relation, which involves a logarithmic relation, we were using M_halo as unit mass by setting UNIT_DENSITY. This is because the natural Mass sizes to scale would be for gas
This is not so common. However, the more common would be power (many astrophysical quantities are expressed in log10) which I also now added.

yt_idefix/data_structures.py
dutta-alankar: I wanted to use the minimum number of files as possible. Since, ```grid.out``` has a lot of information on the grid, I thought using ```grid.out``` for parsing would exhaustively cover almost everything needed by yt. Let me know if this is a good idea and is worth implementing in the parent classes as well.

dutta-alankar: > Agreed. But the parent class needs to be explicitly called

Yes, right now such an explicit call would most likely generate errors.

yt_idefix/data_structures.py
dutta-alankar: I have made the function static

yt_idefix/data_structures.py
dutta-alankar: Done!

yt_idefix/data_structures.py
dutta-alankar: Would you please put this as a docstring wherever you deem appropriate? I'll commit that.

yt_idefix/data_structures.py
dutta-alankar: I see that ```self.parameter_filename``` is used in many parts of the code and I'm a bit confused on where to replace ```self.parameter_filename``` with ```self.filename```.
I have used the ```self.directory``` shortcut as you suggested.

yt_idefix/data_structures.py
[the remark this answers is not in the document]
dutta-alankar: > What I meant was, `split()` is returning a list of str, and you're taking only the third element, but I have no idea how many of them are expected, and what the others are expected to be. Ideally I would like the answers to my questions to be expressed in the code itself, or by a comment.

As mentioned before, the following are entries from ```grid.out``` that are parsed:
```
# X1: [ 0.040000,  0.500000], 400 point(s), 2 ghosts
# X2: [ 0.000000,  6.283185], 400 point(s), 2 ghosts
```
After removing '[', ']' and ',' then the split operation results in an array like:
```
['#', 'X1:', '0.040000', '0.500000', '400', 'point(s)', '2', 'ghosts']
```
Starting the count from 0,
element 2: left edge of the entire domain
element 3: right edge of the entire domain
element 4: number of discrete grid cells

Each such array/line in ```grid.out``` is a dimension.

dutta-alankar:
> I've looked at type-checking errors and my conclusion is that they are triggered because `PlutoXdmfDataset` inherits directly from `PlutoVtkDataset`, while it should inherit from a more generic `PlutoDataset` class that is currently missing; anyway, I can handle that myself later, let's put it aside for the time being. I'll do a first round of review now.

I created a superclass ```PlutoStaticDataset``` (```PlutoDataset``` already exists probably the Chombo Pluto AMR frontend). Now the ```PlutoVtkDataset``` and ```PlutoXdmfDataset``` inherits from ```PlutoStaticDataset```. This is very preliminary and probably a lot of you input is needed to make it as perfect as possible.

yt_idefix/data_structures.py
xshaokun: Thanks, I got it!

xshaokun:
@dutta-alankar, I noticed that `pre-commit.ci` keeps committing to fixes your code. I suggest that you can just install `pre-commit` locally with pip so that it will check and fix your code style before you commit. It's following yt's guideline, see https://yt-project.org/docs/dev/developing/developing.html#coding-style-guide

Maybe it's not forcible but it'll make the commit history more concise :)

neutrinoceros:
I intend to rewrite the history of this branch ultimately so it won't matter too much, but yes it's best if you can use pre-commit locally, especially when many iterations are needed.
I also noticed that this PR is using the `main` branch from your fork, and that's going to be a problem. Let's resolve the important points here, then I will guide you to create another branch (and PR) with your changes, so that you can also integrate other changes (from #204 for instance)

dutta-alankar:
> @dutta-alankar, I noticed that `pre-commit.ci` keeps committing to fixes your code. I suggest that you can just install `pre-commit` locally with pip so that it will check and fix your code style before you commit. It's following yt's guideline, see https://yt-project.org/docs/dev/developing/developing.html#coding-style-guide
> 
> Maybe it's not forcible but it'll make the commit history more concise :)

Sure! I'll use `pre-commit` locally from now on. Actually, I'm very new to yt and also using GitHub. I wasn't aware of this pre-commit feature.

dutta-alankar:
> I intend to rewrite the history of this branch ultimately so it won't matter too much, but yes it's best if you can use pre-commit locally, especially when many iterations are needed. I also noticed that this PR is using the `main` branch from your fork, and that's going to be a problem. Let's resolve the important points here, then I will guide you to create another branch (and PR) with your changes, so that you can also integrate other changes (from #204 for instance)

Yes sure! Sounds fine.

neutrinoceros:
ok so now that #204 is in, here's what you need to do:
```shell
git checkout checkout main
git branch xdmf # create a new branch identical to your "main"

git checkout 9aa8266 # last common ancestor from your main and mine
git push --force # remove all commits from your main branch that are not on mine
```
visit https://github.com/dutta-alankar/yt_idefix
click the "sync fork" button (it'll synchronise your main branch back with mine)
Screenshot 2023-02-25 at 09 35 29

```shell
git checkout main
git pull
git checkout xdmf
git rebase main
git push --set-upstream origin xdmf
```

and open a new PR with the new branch !

neutrinoceros:
Sorry this is so complicated, I should have paid attention to the name of your branch from the beginning.
To sum this up the process, the lesson is: always create a branch before adding new commits, even on your fork. Your main branch should never diverge from the main repo, it can only fall behind, so it's always to synchronise back by just pulling.

dutta-alankar:
> Sorry this is so complicated, I should have paid attention to the name of your branch from the beginning. To sum this up the process, the lesson is: always create a branch before adding new commits, even on your fork. Your main branch should never diverge from the main repo, it can only fall behind, so it's always to synchronise back by just pulling.

I'll wait or a while for @xshaokun 's reply on a minor change I suggested. Once it gets merged to your repo, I'll do the branch creation.

dutta-alankar:
`git push --force` gives me error
```
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git push --force
fatal: You are not currently on a branch.
To push the history leading to the current (detached HEAD)
state now, use

    git push origin HEAD:<name-of-remote-branch>


```

neutrinoceros:
Sorry I never have to do this so I'm a little rusty ^^

So indeed I shouldn't have included this line
```shell
git checkout 9aa8066
```
so let's get back to the previous state with
```shell
git checkout main
```
and do things correctly this time:
```shell
git reset --hard 9aa8066
```
then you should be back on track and you may continue with
```shell
git push --force
...
```

yt_idefix/misc.py
dutta-alankar: Refactored the code to exploit inheritance.

yt_idefix/data_structures.py
dutta-alankar: I think this has now been addressed.

yt_idefix/data_structures.py
dutta-alankar: Addressed!

yt_idefix/data_structures.py
xshaokun: > I wanted to use the minimum number of files as possible.

I'm totally agreed! I just gave you some suggestion there. However that means you need to do a lot of work and much of existing work will be dropped. Please think about that.

PR #204, opened 2023-02-24
fix setting code units in PlutoDataset

xshaokun:
Reminded by the discussion with @dutta-alankar in #202, some bugs need to be fixed.

- [x] adopt the correct default code unit values in Pluto
- [x] add the parsers for `log` and `log10` in unit definitions

PR #206, opened 2023-02-25
fix repeatedly replace 'log' in unit definitions

xshaokun:
Fix the bug introduced in #204. A test should be added for this.


yt_idefix/data_structures.py
```diff
@@ -598,7 +598,6 @@ def _parse_definitions_header(self) -> None:
                 expr = re.sub(r"UNIT_(\w+)", self._get_unit, expr)
                 expr = re.sub(r"sqrt", "np.sqrt", expr)
                 expr = re.sub(r"log", "np.log", expr)
-                expr = re.sub(r"log10", "np.log10", expr)
```
neutrinoceros, line 601: maybe the issue is that we're replacing "log", *then* "log10" ? in which case the comment on line 596 should be updated too ?

neutrinoceros: to be clear, I think inverting the order of operations should do the trick

xshaokun:
I added log and log10 in a lazy way. But I think it's just for test, so the simulation results are not so much important here.

yt_idefix/data_structures.py
xshaokun: As @dutta-alankar said [here](https://github.com/neutrinoceros/yt_idefix/pull/204#discussion_r1117888728), actually `re.sub(r"log", "np.log", expr)` is able to deal with both "log" and "log10", and even "log2". Because log10 contains the string "log", right?

neutrinoceros: Oh that's right, good call

neutrinoceros:
I may need to revert this because the following file wasn't handled properly using git-lfs
```
tests/data/pluto_sod/data.0001.vtk
```

PR #207, opened 2023-02-25
BUG: convert expected current time to unyt_quantity

xshaokun:
This PR should be merged before #206


xshaokun:
Oh I'm sorry, it's not a bug. I just gave the wrong type of value to `current_time` in meta.yaml.

PR #209, opened 2023-02-25
PLUTO xdmf dump support

dutta-alankar:
As discussed in the previous PR #202 
This is a new branch xdmf that can be merged!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.egg-info
 **/__pycache__
 dist/*
+venv/*
--- a/tests/test_xdmf.py
+++ b/tests/test_xdmf.py
@@ -1,0 +1,41 @@
+from pathlib import Path
+
+import pytest
+
+import yt
+from yt_idefix.api import PlutoXdmfDataset
+
+pytest.importorskip("h5py")
+
+
+HERE = Path(__file__).parent
+DATADIR = HERE / "data"
+ds_path_plutoXDMF = DATADIR.joinpath("pluto_isentropic_vortex", "data.0010.flt.h5")
+
+
+def test_class_validation(xdmf_file):
+    file = xdmf_file
+    assert PlutoXdmfDataset._is_valid(str(file["path"]))
+
+
+def test_load_class(xdmf_file):
+    file = xdmf_file
+    PlutoXdmfDataset(str(file["path"]))
+
+
+# TODO: make this a pytest-mpl test
+def test_slice_plot(xdmf_file):
+    file = xdmf_file
+    ds = yt.load(file["path"], geometry=file["geometry"], unit_system="code")
+    yt.SlicePlot(ds, normal=(0, 0, 1), fields=("gas", "density"))
+
+
+def test_projection_plot(xdmf_file):
+    file = xdmf_file
+    ds = yt.load(file["path"], geometry=file["geometry"], unit_system="code")
+    yt.ProjectionPlot(ds, normal=(0, 0, 1), fields=("gas", "density"))
+
+
+def test_load_magic(xdmf_file):
+    ds = yt.load(xdmf_file["path"], geometry=xdmf_file["geometry"])
+    assert isinstance(ds, PlutoXdmfDataset)
--- a/yt_idefix/_io/commons.py
+++ b/yt_idefix/_io/commons.py
@@ -47,6 +47,56 @@ class Coordinates(NamedTuple):
         return Coordinates(arrs[0], arrs[1], arrs[2], self.array_shape)
 
 
+def get_native_coordinates_from_cartesian(
+    xcart: np.ndarray,
+    ycart: np.ndarray,
+    zcart: np.ndarray,
+    geometry: str,
+) -> list[np.ndarray]:
+    shape = Shape(*xcart.shape)
+
+    # Reconstruct the polar coordinate system
+    if geometry == "polar":
+        # on disk coordinates are cell face coordinates.
+        # We need to convert from cartesian to polar,
+        # but no interpolation is needed.
+        r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2)
+        theta = np.unwrap(np.arctan2(ycart[0, :, 0], xcart[0, :, 0]))
+        z = zcart[0, 0, :]
+
+        coords = [r, theta, z]
+    elif geometry == "spherical":
+        # Reconstruct the spherical coordinate system
+        if shape.n3 == 1:
+            r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2)
+            phi = np.unwrap(
+                np.arctan2(zcart[0, shape.n2 // 2, :], xcart[0, shape.n2 // 2, :])
+            )
+            theta = np.arccos(
+                ycart[0, :, 0] / np.sqrt(xcart[0, :, 0] ** 2 + ycart[0, :, 0] ** 2)
+            )
+        else:
+            r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2 + zcart[:, 0, 0] ** 2)
+            phi = np.unwrap(
+                np.arctan2(
+                    ycart[shape.n1 // 2, shape.n2 // 2, :],
+                    xcart[shape.n1 // 2, shape.n2 // 2, :],
+                )
+            )
+            theta = np.arccos(
+                zcart[0, :, 0]
+                / np.sqrt(
+                    xcart[0, :, 0] ** 2 + ycart[0, :, 0] ** 2 + zcart[0, :, 0] ** 2
+                )
+            )
+        coords = [r, theta, phi]
+    else:
+        raise NotImplementedError(
+            f"This kind of geometry: {geometry} is not supported yet!"
+        )
+    return coords
+
+
 # map field name to numpy array init data:
 # precision (-> datatype), dimensionality, [nx, ny, nz]
 # the np.ndarray is assumed to contain *dim* elements
--- a/yt_idefix/_io/h5_io.py
+++ b/yt_idefix/_io/h5_io.py
@@ -1,0 +1,89 @@
+from __future__ import annotations
+
+import os
+
+import numpy as np
+
+from yt.utilities.on_demand_imports import _h5py as h5py
+
+from .commons import Coordinates, Shape, get_native_coordinates_from_cartesian
+
+KNOWN_GEOMETRIES: dict[int, str] = {
+    0: "cartesian",
+    1: "polar",
+    2: "spherical",
+    3: "cylindrical",
+}
+
+
+def read_grid_coordinates(
+    filename: str | os.PathLike[str],
+    *,
+    geometry: str | None = None,
+) -> Coordinates:
+    # Return cell edges coordinates
+    fh = h5py.File(filename, "r")
+    if geometry not in (valid_geometries := tuple(KNOWN_GEOMETRIES.values())):
+        raise ValueError(
+            f"Got unknown geometry {geometry!r}, expected one of {valid_geometries}"
+        )
+
+    nodesX = fh["/node_coords/X"].astype("=f8")
+    nodesY = fh["/node_coords/Y"].astype("=f8")
+    nodesZ = fh["/node_coords/Z"].astype("=f8")
+    dimensions = len(np.array(nodesX).shape)
+    shape = Shape(
+        *np.array(nodesX).shape
+    )  # this is reversed compared the vtk implementation in vtk_io.py
+    coords: list[np.ndarray] = []
+    # now assuming that fh is positioned at the end of metadata
+    if geometry in ("cartesian", "cylindrical") or (
+        geometry in ("polar", "spherical") and dimensions == 1
+    ):
+        if dimensions == 1:
+            nodesX = np.array(nodesX)
+            # Default is assumed and not parsed from pluto.ini/grid.out (not present in data)
+            nodesY = np.array([0.0, 1.0])
+            nodesZ = np.array([0.0, 1.0])
+            if geometry == "spherical":
+                if np.fabs(np.max(nodesX) - np.min(nodesX)) < 1e-8:
+                    nodesX = fh["/cell_coords/X"].astype("=f8")
+                    nodesX = np.hstack(([nodesX[1] - nodesX[0]], np.array(nodesX)))
+                    nodesX = np.array(nodesX) / np.sin(0.5)
+            elif geometry == "polar":
+                if np.fabs(np.max(nodesX) - np.min(nodesX)) < 1e-8:
+                    nodesX = fh["/cell_coords/X"].astype("=f8")
+                    nodesX = np.hstack(([nodesX[1] - nodesX[0]], np.array(nodesX)))
+                    nodesX = np.array(nodesX) / np.cos(0.5)
+            array_shape = Shape(shape[0], 1, 1).to_cell_centered()
+        elif dimensions == 2:
+            nodesX = np.array(nodesX[0, :])
+            nodesY = np.array(nodesY[:, 0])
+            # Default is assumed and not parsed from pluto.ini/grid.out (not present in data)
+            if geometry == "cartesian":
+                nodesZ = np.array([0.0, 1.0])
+            else:
+                nodesZ = np.array([0.0, 2 * np.pi])
+            array_shape = Shape(*reversed(shape[:-1])).to_cell_centered()
+        else:
+            nodesX = np.array(nodesX)[0, 0, :]
+            nodesY = np.array(nodesY)[0, :, 0]
+            nodesZ = np.array(nodesZ)[:, 0, 0]
+            array_shape = Shape(*reversed(shape)).to_cell_centered()
+        coords = [nodesX, nodesY, nodesZ]
+    elif geometry in ("polar", "spherical") and dimensions > 1:
+        if dimensions == 2:
+            nodesX = np.array([nodesX])
+            nodesY = np.array([nodesY])
+            nodesZ = np.array([nodesZ])
+            array_shape = Shape(*reversed(shape[:-1])).to_cell_centered()
+        else:
+            array_shape = Shape(*reversed(shape)).to_cell_centered()
+
+        xcart = np.transpose(nodesX, axes=(2, 1, 0))
+        ycart = np.transpose(nodesY, axes=(2, 1, 0))
+        zcart = np.transpose(nodesZ, axes=(2, 1, 0))
+
+        coords = get_native_coordinates_from_cartesian(xcart, ycart, zcart, geometry)
+    fh.close()
+    return Coordinates(coords[0], coords[1], coords[2], array_shape)
--- a/yt_idefix/api.py
+++ b/yt_idefix/api.py
@@ -3,4 +3,5 @@ from .data_structures import (
     IdefixDmpDataset,
     IdefixVtkDataset,
     PlutoVtkDataset,
+    PlutoXdmfDataset,
 )
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -20,12 +20,20 @@ from yt.geometry.grid_geometry_handler import GridIndex
 from yt.utilities.lib.misc_utilities import (  # type: ignore [import]
     _obtain_coords_and_widths,
 )
+from yt.utilities.on_demand_imports import _h5py as h5py
+from yt_idefix._backports import removesuffix
 from yt_idefix._typing import UnitLike
 
-from ._io import C_io, dmp_io, vtk_io
+from ._io import C_io, dmp_io, h5_io, vtk_io
 from ._io.commons import IdefixFieldProperties, IdefixMetadata
 from .definitions import _PlutoBaseUnits, pluto_def_constants
-from .fields import BaseVtkFields, IdefixDmpFields, IdefixVtkFields, PlutoVtkFields
+from .fields import (
+    BaseVtkFields,
+    IdefixDmpFields,
+    IdefixVtkFields,
+    PlutoVtkFields,
+    PlutoXdmfFields,
+)
 
 # import IO classes to ensure they are properly registered,
 # even though we don't call them directly
@@ -87,8 +95,6 @@ class GoodBoyHierarchy(GridIndex, ABC):
         # Idefix/Pluto are not AMR
         self.grid_levels[0][0] = 0
         self.min_level = self.max_level = 0
-
-        self._field_offsets = self._get_field_offset_index()
 
     def _populate_grid_objects(self):
         # the minimal form of this method is
@@ -240,6 +246,61 @@ class IdefixDmpHierarchy(FieldOffsetHierarchy):
             fdata["x2"] * length_unit,
             fdata["x3"] * length_unit,
         )
+
+
+class PlutoXdmfHierarchy(GoodBoyHierarchy):
+    def _detect_output_fields(self):
+        with h5py.File(self.index_filename, mode="r") as h5f:
+            root = list(h5f.keys())[0]
+            self.field_list = [
+                (self.ds._dataset_type, str(k)) for k in list(h5f[f"{root}/vars/"])
+            ]
+
+    @cached_property
+    def _cell_widths(self) -> tuple[XSpans, YSpans, ZSpans]:
+        cell_edges = h5_io.read_grid_coordinates(
+            self.index_filename, geometry=self.ds.geometry
+        )
+
+        dims = self.ds.domain_dimensions
+        length_unit = self.ds.quan(1, "code_length")
+
+        cell_widths: tuple[XSpans, YSpans, ZSpans]
+        cell_widths = (
+            np.empty(max(dims[0], 2), dtype="float64") * length_unit,
+            np.empty(max(dims[1], 2), dtype="float64") * length_unit,
+            np.empty(max(dims[2], 2), dtype="float64") * length_unit,
+        )
+
+        for idir, edges in enumerate(cell_edges[:3]):
+            if dims[idir] > 1:
+                cell_widths[idir][:] = np.ediff1d(edges)
+            else:
+                cell_widths[idir][:] = self.ds.domain_width[idir]
+        return cell_widths
+
+    @cached_property
+    def _cell_centers(self) -> tuple[XCoords, YCoords, ZCoords]:
+        cell_edges = h5_io.read_grid_coordinates(
+            self.index_filename, geometry=self.ds.geometry
+        )
+
+        dims = self.ds.domain_dimensions
+        length_unit = self.ds.quan(1, "code_length")
+
+        cell_centers: tuple[XCoords, YCoords, ZCoords]
+        cell_centers = (
+            np.empty(max(dims[0], 2), dtype="float64") * length_unit,
+            np.empty(max(dims[1], 2), dtype="float64") * length_unit,
+            np.empty(max(dims[2], 2), dtype="float64") * length_unit,
+        )
+
+        for idir, edges in enumerate(cell_edges[:3]):
+            if dims[idir] > 1:
+                cell_centers[idir][:] = 0.5 * (edges[1:] + edges[:-1])
+            else:
+                cell_centers[idir][:] = edges[0]
+        return cell_centers
 
 
 class GoodboyDataset(Dataset, ABC):
@@ -765,3 +826,100 @@ class PlutoVtkDataset(VtkMixin, StaticPlutoDataset):
 
     def _get_log_file(self) -> str:
         return os.path.join(self.directory, "vtk.out")
+
+
+class PlutoXdmfDataset(StaticPlutoDataset):
+    _dataset_type = "pluto-xdmf"
+    _index_class = PlutoXdmfHierarchy
+    _field_info_class = PlutoXdmfFields
+
+    def _get_log_file(self) -> str:
+        if (suffix := re.search(r"(flt|dbl)\.h5$", self.filename)) is not None:
+            return os.path.join(self.directory, f"{suffix.group()}.out")
+        else:
+            raise RuntimeError(
+                f"Failed to detect log file associated with {self.filename}"
+            )
+
+    def _parse_parameter_file(self):
+        """
+        Filenames are data.<snapnum>.<dbl/flt>.h5
+        <snapnum> needs to be parse from the filename.
+        <snapnum> is the corresponding entry in the <dbl/flt>.h5.out file
+        Example <dbl/flt>.h5.out file:
+            0 0.000000e+00 1.000000e-04 0 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            1 2.498181e+00 3.500985e-03 747 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            2 4.998045e+00 3.400969e-03 1458 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            3 7.497932e+00 3.386245e-03 2186 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+
+        One of these lines is parsed to count the number of passive tracer fields in the data dump.
+        """
+        super()._parse_parameter_file()
+
+        # parse the grid
+        coords = h5_io.read_grid_coordinates(
+            self.filename, geometry=self.parameters["definitions"]["geometry"]
+        )
+
+        self.domain_dimensions = np.array(coords.array_shape)
+        self.dimensionality = np.count_nonzero(self.domain_dimensions - 1)
+
+        dle = np.array([arr.min() for arr in coords.arrays], dtype="float64")
+        dre = np.array([arr.max() for arr in coords.arrays], dtype="float64")
+
+        # temporary hack to prevent 0-width dimensions for 2D data
+        dre = np.where(dre == dle, dle + 1, dre)
+        self.domain_left_edge = dle
+        self.domain_right_edge = dre
+
+        self._periodicity = (True, True, True)
+
+    def _count_tracers(self) -> int:
+        with open(self._get_log_file()) as fh:
+            txt = fh.readlines()
+            if (match := re.search(r"\d{4}", self.filename)) is not None:
+                entry = int(match.group())
+            else:
+                raise RuntimeError(f"Failed to parse entry number from {self.filename}")
+
+            count = 0
+            # Passive tracer names in PLUTO are tr1, tr2, tr3 and so on by default
+            while f"tr{count + 1}" in txt[entry].split():
+                count += 1
+
+        return count
+
+    def _read_data_header(self) -> str:
+        grid_file = os.path.join(self.directory, "grid.out")
+        if not os.path.isfile(grid_file):
+            return ""
+
+        with open(grid_file) as fh:
+            body = fh.read()
+
+        if (res := re.match(r"(#.*\n)+", body)) is not None:
+            return res.group()
+        else:
+            return ""
+
+    @classmethod
+    def _is_valid(cls, filename: str, *args, **kwargs) -> bool:
+        if not (
+            filename.endswith((".dbl.h5", ".flt.h5"))
+            and os.path.isfile(removesuffix(filename, ".h5") + ".xmf")
+            and os.path.isfile(
+                os.path.join(
+                    os.path.dirname(filename), os.path.basename(filename[-6:]) + ".out"
+                )
+            )
+        ):
+            return False
+
+        try:
+            fileh = h5py.File(filename, mode="r")
+        except (ImportError, OSError):
+            return False
+        else:
+            entries = list(fileh.keys())
+            fileh.close()
+            return "cell_coords" in entries and "node_coords" in entries
--- a/yt_idefix/fields.py
+++ b/yt_idefix/fields.py
@@ -2,7 +2,12 @@ from yt.fields.field_info_container import FieldInfoContainer
 from yt.fields.magnetic_field import setup_magnetic_field_aliases
 
 
-class BaseVtkFields(FieldInfoContainer):
+class BaseFields(FieldInfoContainer):
+    # This class can be used later to add functions and attributes common to all fields
+    pass
+
+
+class BaseVtkFields(BaseFields):
     known_other_fields = (
         ("RHO", ("code_mass / code_length**3", ["density"], None)),  # type: ignore
         ("VX1", ("code_length / code_time", ["velocity_x"], None)),
@@ -27,7 +32,7 @@ class PlutoVtkFields(BaseVtkFields):
         setup_magnetic_field_aliases(self, "pluto-vtk", [f"BX{idir}" for idir in "123"])
 
 
-class IdefixDmpFields(FieldInfoContainer):
+class IdefixDmpFields(BaseFields):
     known_other_fields = (
         ("Vc-RHO", ("code_mass / code_length**3", ["density"], None)),  # type: ignore
         ("Vc-VX1", ("code_length / code_time", ["velocity_x"], None)),
@@ -47,3 +52,87 @@ class IdefixDmpFields(FieldInfoContainer):
         setup_magnetic_field_aliases(
             self, "idefix-dmp", [f"Vc-BX{idir}" for idir in "123"]
         )
+
+
+class PlutoXdmfFields(BaseFields):
+    _pres_units = "code_pressure"
+    _erg_units = "code_mass * (code_length/code_time)**2"
+    _rho_units = "code_mass / code_length**3"
+    _mom_units = "code_mass / code_length**2 / code_time"
+    _vel_units = "code_length/code_time"
+
+    known_other_fields = (
+        # Each entry here is of the form
+        # ( "name", ("units", ["fields", "to", "alias"], # "display_name")),
+        ("rho", (_rho_units, ["density", "rho"], None)),
+        ("vx1", (_vel_units, ["vel_x"], None)),
+        ("vx2", (_vel_units, ["vel_y"], None)),
+        ("vx3", (_vel_units, ["vel_z"], None)),
+        ("prs", (_pres_units, ["prs", "pres", "pressure"], None)),
+        ("Bx1", ("code_magnetic", [], None)),
+        ("Bx2", ("code_magnetic", [], None)),
+        ("Bx3", ("code_magnetic", [], None)),
+    )
+
+    known_particle_fields = ()
+
+    def setup_fluid_fields(self):
+        unit_system = self.ds.unit_system
+
+        setup_magnetic_field_aliases(
+            self, self.ds._dataset_type, [f"Bx{idir}" for idir in "123"]
+        )
+
+        # Add tracer fields
+        for i in range(1, self.ds._count_tracers() + 1):
+            if (self.ds._dataset_type, f"tr{i}") in self.field_list:
+                self.add_output_field(
+                    (self.ds._dataset_type, f"tr{i}"),
+                    sampling_type="cell",
+                    units="",
+                )
+                self.alias(
+                    ("gas", f"tracer_{i}"),
+                    (self.ds._dataset_type, f"tr{i}"),
+                    units="",
+                )
+        # Add temperature field, these are commonly defined as either "Temp" or "Temperature" by users in PLUTO
+        if (self.ds._dataset_type, "Temp") in self.field_list:
+            self.add_output_field(
+                (self.ds._dataset_type, "Temp"),
+                sampling_type="cell",
+                units=unit_system["temperature"],
+            )
+            self.alias(
+                ("gas", "temperature"),
+                (self.ds._dataset_type, "Temp"),
+                units=unit_system["temperature"],
+            )
+        elif (self.ds._dataset_type, "Temperature") in self.field_list:
+            self.add_output_field(
+                (self.ds._dataset_type, "Temperature"),
+                sampling_type="cell",
+                units=unit_system["temperature"],
+            )
+            self.alias(
+                ("gas", "temperature"),
+                (self.ds._dataset_type, "Temperature"),
+                units=unit_system["temperature"],
+            )
+        # Optional particle number density field in PLUTO
+        if (self.ds._dataset_type, "ndens") in self.field_list:
+            self.add_output_field(
+                (self.ds._dataset_type, "ndens"),
+                sampling_type="cell",
+                units=unit_system["length"] ** -3,
+            )
+            self.alias(
+                ("gas", "number_density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
+
+    def setup_particle_fields(self, ptype):
+        super().setup_particle_fields(ptype)
+        # This will get called for every particle type.
+        # {TODO} Starting with version 4.4 PLUTO has particle support and this function needs an implementation in future
--- a/yt_idefix/io.py
+++ b/yt_idefix/io.py
@@ -1,9 +1,11 @@
+import re
 from abc import ABC, abstractmethod
 from typing import BinaryIO, Tuple, cast
 
 import numpy as np
 
 from yt.utilities.io_handler import BaseIOHandler, BaseParticleIOHandler
+from yt.utilities.on_demand_imports import _h5py as h5py
 
 from ._io import dmp_io, vtk_io
 
@@ -80,3 +82,54 @@ class IdefixDmpIO(SingleGridIO, BaseParticleIOHandler):
     def _read_particle_fields(self, chunks, ptf, selector):
         # idefix doesn't have particles (yet)
         raise NotImplementedError("Particles are not currently supported for Idefix")
+
+
+class PlutoXdmfIOHandler(BaseIOHandler):
+    _particle_reader = False
+    _dataset_type = "pluto-xdmf"
+
+    def _read_particle_coords(self, chunks, ptf):
+        raise NotImplementedError("Reading of particle fields not yet implemented!")
+
+    def _read_particle_fields(self, chunks, ptf, selector):
+        raise NotImplementedError("Reading of particle fields not yet implemented!")
+
+    def _read_fluid_selection(self, chunks, selector, fields, size):
+        """
+        Filenames are data.<snapnum>.<dbl/flt>.h5
+        <snapnum> needs to be parse from the filename.
+        <snapnum> is the corresponding entry in the <dbl/flt>.h5.out file
+        Example <dbl/flt>.h5.out file:
+            0 0.000000e+00 1.000000e-04 0 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            1 2.498181e+00 3.500985e-03 747 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            2 4.998045e+00 3.400969e-03 1458 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            3 7.497932e+00 3.386245e-03 2186 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+        """
+        if (match := re.search(r"\d{4}", self.ds.filename)) is not None:
+            entry = int(match.group())
+        else:
+            raise RuntimeError(f"Failed to parse output number from {self.ds.filename}")
+
+        data = {field: np.empty(size, dtype="float64") for field in fields}
+
+        with h5py.File(self.ds.filename, "r") as fh:
+            ind = 0
+            for chunk in chunks:
+                for grid in chunk.objs:
+                    nd = 0
+                    for field in fields:
+                        ftype, fname = field
+                        position = f"/Timestep_{entry}/vars/{fname}"
+                        field_data = fh[position][:].astype("=f8")
+
+                        while field_data.ndim < 3:
+                            field_data = np.expand_dims(field_data, axis=0)
+
+                        # X3 X2 X1 orderding of fields in PLUTO needs to rearranged to X1 X2 X3 order in yt.
+                        values = np.transpose(field_data, axes=(2, 1, 0))
+                        nd = grid.select(selector, values, data[field], ind)
+                    ind += nd
+        return data
+
+    def _read_chunk_data(self, chunk, fields):
+        raise NotImplementedError("No multipart file dumps happen in PLUTO till date!")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,9 @@ jobs:
         python-version:
         - '3.8'
         - '3.11'
+        install-target:
+        - .
+        - .[HDF5]
 
     runs-on: ${{ matrix.os }}
     steps:
@@ -31,9 +34,11 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Build
+      env:
+        INSTALL_TARGET: ${{ matrix.install-target }}
       run: |
         python -m pip install --upgrade pip
-        python -m pip install .
+        python -m pip install "$INSTALL_TARGET"
         python -m pip install --requirement requirements/tests.txt
 
     - name: run tests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,12 +37,16 @@ dependencies = [
     "yt>=4.1.0",
 ]
 
+[project.optional-dependencies]
+HDF5 = ["h5py>=3.1.0"]
+
 [project.entry-points."yt.frontends"]
 # these entry points require yt 4.2 (unreleased at the time of writing)
 # they are included early for testing purposes
 IdefixDmpkDataset = "yt_idefix.api:IdefixDmpDataset"
 IdefixVtkDataset = "yt_idefix.api:IdefixVtkDataset"
 PlutoVtkDataset = "yt_idefix.api:PlutoVtkDataset"
+PlutoXdmfDataset = "yt_idefix.api:PlutoXdmfDataset"
 
 [project.license]
 text = "GPL-3.0"
--- a/yt_idefix/_io/vtk_io.py
+++ b/yt_idefix/_io/vtk_io.py
@@ -6,7 +6,7 @@ from typing import Any, BinaryIO, Literal, overload
 
 import numpy as np
 
-from .commons import Coordinates, Shape
+from .commons import Coordinates, Shape, get_native_coordinates_from_cartesian
 
 KNOWN_GEOMETRIES: dict[int, str] = {
     0: "cartesian",
@@ -191,50 +191,10 @@ def read_grid_coordinates(
         zcart.shape = rshape
         zcart = zcart.T
 
-        # Reconstruct the polar coordinate system
-        if geometry == "polar":
-            # on disk coordinates are cell face coordinates.
-            # We need to convert from cartesian to polar,
-            # but no interpolation is needed.
-            r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2)
-            theta = np.unwrap(np.arctan2(ycart[0, :, 0], xcart[0, :, 0]))
-            z = zcart[0, 0, :]
+        coords = get_native_coordinates_from_cartesian(xcart, ycart, zcart, geometry)
 
-            data_type = next(fh).decode().split()[0]  # CELL_DATA (NX-1)(NY-1)(NZ-1)
-            next(fh)
-
-            coords = [r, theta, z]
-        else:
-            assert geometry == "spherical"
-            # Reconstruct the spherical coordinate system
-            if shape.n3 == 1:
-                r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2)
-                phi = np.unwrap(
-                    np.arctan2(zcart[0, shape.n2 // 2, :], xcart[0, shape.n2 // 2, :])
-                )
-                theta = np.arccos(
-                    ycart[0, :, 0] / np.sqrt(xcart[0, :, 0] ** 2 + ycart[0, :, 0] ** 2)
-                )
-            else:
-                r = np.sqrt(
-                    xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2 + zcart[:, 0, 0] ** 2
-                )
-                phi = np.unwrap(
-                    np.arctan2(
-                        ycart[shape.n1 // 2, shape.n2 // 2, :],
-                        xcart[shape.n1 // 2, shape.n2 // 2, :],
-                    )
-                )
-                theta = np.arccos(
-                    zcart[0, :, 0]
-                    / np.sqrt(
-                        xcart[0, :, 0] ** 2 + ycart[0, :, 0] ** 2 + zcart[0, :, 0] ** 2
-                    )
-                )
-            coords = [r, theta, phi]
-
-            data_type = next(fh).decode().split()[0]  # CELL_DATA (NX-1)(NY-1)(NZ-1)
-            next(fh)
+        data_type = next(fh).decode().split()[0]  # CELL_DATA (NX-1)(NY-1)(NZ-1)
+        next(fh)
 
         array_shape = shape.to_cell_centered()
         assert data_type == "CELL_DATA"


xshaokun:
@dutta-alankar I just reviewed `hdf5_io.c` in Pluto, and I noticed that output `data.*.h5` contains enough information like other format of outputs. That means you can parse grid data and code version from output file instead of from `grid.out`, and you can get rid of `grid.out`. There're some benefits to do so:

1. Less file dependence. This will make the module easier to use and stable.
2. Make the behaviors of `PlutoXdmfDataset` and `PlutoVtkDataset` more consistent. This will make it easier to develop and  maintain in the future.

I believe that the only difference between `PlutoXdmfDataset` and `PlutoVtkDataset` is the io part, so a `hdf5_io.py` ( refer to `_io/vtk_io.py`) would be able to reduce much work in `data_structure.py`. Namely, you don't have to override so many class methods.

@neutrinoceros What's your opinion?

neutrinoceros:
@xshaokun thank you very much for giving this so much thoughts !
Avoiding effort duplication is the very reason that I encouraged @dutta-alankar to contribute here, so if it makes sense to reduce the size of this patch, I'm all for it !

dutta-alankar:
> @xshaokun thank you very much for giving this so much thoughts ! Avoiding effort duplication is the very reason that I encouraged @dutta-alankar to contribute here, so if it makes sense to reduce the size of this patch, I'm all for it !

Reading directly from hdf5 file has one problem which why I didn't do it. First of all it is given as an array of (nx3,nx2,nx1) rather than individual 1d arrays along 3 directions like in `grid.out`. Also for non-cartesian geometry, it does conversion of all grid points to cartesian and then dumps the cartesian points. Using this then, requires setting up unstructured mesh with mesh conectivities specified. I have done this in C in a different context, but this is not straightforward to implement in yt.

neutrinoceros:
Vtk also pose this limitation, and we already have a lot of geometry conversion code in `vtk_io.py`. Maybe it would make sense to recycle it.

dutta-alankar:
> Vtk also pose this limitation, and we already have a lot of geometry conversion code in `vtk_io.py`. Maybe it would make sense to recycle it.

In that case perhaps, it is worth checking out. I'll go through the code then!

neutrinoceros:
It's in `yt_idefix._io.vtk_io.read_grid_coordinates`

dutta-alankar:
I have followed suggestions from @xshaokun and @neutrinoceros and now the grid reading is done using the pre-existing code in the vtk reading functions. Please provide your comments and suggestions regarding this. Now `grid.out`
is only used for reading optionally the PLUTO code version.

> > @xshaokun thank you very much for giving this so much thoughts ! Avoiding effort duplication is the very reason that I encouraged @dutta-alankar to contribute here, so if it makes sense to reduce the size of this patch, I'm all for it !
> 
> Reading directly from hdf5 file has problems which why I didn't do it. First of all, it is given as an array of (nx3,nx2,nx1) meshgrid rather than individual 1d arrays along 3 directions like in `grid.out`. Also, for non-cartesian geometry, it does conversion of all grid points to cartesian and then dumps the cartesian points. Using this then, requires setting up unstructured mesh with mesh conectivities specified. I have done this in C in a different context, but this is not straightforward to implement in yt.

neutrinoceros:
At this point, I think we need to prioritise adding a test case.
Caution is needed to add any heavy file to the repo using git-lfs, and I won't be able to assist with the details of this for a couple days, so for now I advise you work on it locally.
Here's what you can do in the mean time

- prepare at least one (maybe two) small (~1Mb or less) dataset with `pluto.ini`, `definitions.h` and a `meta.yaml` file in the style of existing tests *but don't commit the datafiles to your branch just yet*, and make sure it's working as intended with minimal manual tests.
- define a new pytest fixture (`pluto_hdf5_file`) in `tests/confest.py` in the style of `pluto_vtk_file`
- add a module `tests/test_hdf5.py` and define a variant of `test_vtk.test_parse_pluto_metadata` there (using `pluto_hdf5_file` instead of `pluto_vtk_file`)

committing and pushing text files is always okay. Do not hesitate with any questions

dutta-alankar:
> At this point, I think we need to prioritise adding a test case. Caution is needed to add any heavy file to the repo using git-lfs, and I won't be able to assist with the details of this for a couple days, so for now I advise you work on it locally. Here's what you can do in the mean time
> 
> * prepare at least one (maybe two) small (~1Mb or less) dataset with `pluto.ini`, `definitions.h` and a `meta.yaml` file in the style of existing tests _but don't commit the datafiles to your branch just yet_, and make sure it's working as intended with minimal manual tests.
> * define a new pytest fixture (`pluto_hdf5_file`) in `tests/confest.py` in the style of `pluto_vtk_file`
> * add a module `tests/test_hdf5.py` and define a variant of `test_vtk.test_parse_pluto_metadata` there (using `pluto_hdf5_file` instead of `pluto_vtk_file`)
> 
> committing and pushing text files is always okay. Do not hesitate with any questions

Noted! I'll get to it rightaway.

dutta-alankar:
> > At this point, I think we need to prioritise adding a test case. Caution is needed to add any heavy file to the repo using git-lfs, and I won't be able to assist with the details of this for a couple days, so for now I advise you work on it locally. Here's what you can do in the mean time
> > 
> > * prepare at least one (maybe two) small (~1Mb or less) dataset with `pluto.ini`, `definitions.h` and a `meta.yaml` file in the style of existing tests _but don't commit the datafiles to your branch just yet_, and make sure it's working as intended with minimal manual tests.
> > * define a new pytest fixture (`pluto_hdf5_file`) in `tests/confest.py` in the style of `pluto_vtk_file`
> > * add a module `tests/test_hdf5.py` and define a variant of `test_vtk.test_parse_pluto_metadata` there (using `pluto_hdf5_file` instead of `pluto_vtk_file`)
> > 
> > committing and pushing text files is always okay. Do not hesitate with any questions
> 
> Noted! I'll get to it rightaway.

Also please review `_io/vtk_io.py` and `_io/commons.py`. I have changed these codes a bit to use the geometric stuffs also in xdmf io as well.

xshaokun:
@dutta-alankar The structure looks much better than before! I'll help to review more details later!

dutta-alankar:
Seems like the type check has failed but the complaining lines are right and the code works. @neutrinoceros Can you help?

neutrinoceros:
Sorry I don't have time to help here until tomorrow (24h from now). I'll try to check in and provide as much insight as possible then.

dutta-alankar:
> Sorry I don't have time to help here until tomorrow (24h from now). I'll try to check in and provide as much insight as possible then.

No hurries and no problem at all.

yt_idefix/data_structures.py
```diff
@@ -162,13 +169,17 @@ def _icoords_to_fcoords(
...
+    _io_type = vtk_io
+
     def _get_field_offset_index(self) -> dict[str, int]:
         return self.ds._field_offset_index
 
     @cached_property
     def _cell_widths(self) -> tuple[XSpans, YSpans, ZSpans]:
         with open(self.index_filename, "rb") as fh:
-            cell_edges = vtk_io.read_grid_coordinates(fh, geometry=self.ds.geometry)
+            cell_edges = self._io_type.read_grid_coordinates(
```
xshaokun, line 180: I noticed `vtk_io.read_grid_coordinates` accepts `BinaryIO` as the first argument while `h5_io.read_grid_coordinates` accepts str or PathLike. That means you cannot generalize this implement for now. 

I believe it will raise error when `_cell_widths` and `_cell_centers` below are called.

xshaokun (changes requested):
Sorry for late. It's my first round of review, mainly on the structure. Beside, I gave my resolution to your type check error. Hope it will work :)

yt_idefix/data_structures.py
dutta-alankar: > I noticed `vtk_io.read_grid_coordinates` accepts `BinaryIO` as the first argument while `h5_io.read_grid_coordinates` accepts str or PathLike. That means you cannot generalize this implement for now.
> 
> I believe it will raise error when `_cell_widths` and `_cell_centers` below are called.

This is way for now I copied most of the code. This isn't ideal but works out for initial testing.

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,243 @@
...
+        parse_shape(next(fh).decode(), metadata)
+
+    elif line.startswith("DIMENSIONS"):
+        parse_shape(line, metadata)
+
+    else:
+        raise RuntimeError(f"Failed to parse {line!r}")
+
+    return metadata
+
```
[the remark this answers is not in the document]
dutta-alankar: Yes! Worth trying.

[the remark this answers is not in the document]
dutta-alankar: There is no metadata reading in hdf5. So, these functions are not used by xdmf files.

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,243 @@
...
+                    nodesX = np.array(nodesX) / np.cos(0.5)
+            array_shape = Shape(shape[0], 1, 1).to_cell_centered()
+        elif dimensions == 2:
+            nodesX = np.array(nodesX[0, :])
+            nodesY = np.array(nodesY[:, 0])
+            if geometry == "cartesian":
+                nodesZ = np.array([0.0, 1.0])
+            else:
+                nodesZ = np.array([0.0, 2 * np.pi])
+            array_shape = Shape(*reversed(shape[:-1]), 1).to_cell_centered()
```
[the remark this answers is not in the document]
dutta-alankar: Seems like this still fails the check.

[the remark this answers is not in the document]
xshaokun: Well, I tested your branch locally. And I found it works in this way:

```suggestion
            array_shape = Shape(*reversed(shape[:-1])).to_cell_centered()
```

Three position arguments in Shape have default values, so don't need to give them all.

FYI, if you want to type check locally. You should install `mypy` and run `mypy .` in your `yt_idefix` directory.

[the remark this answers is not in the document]
xshaokun: Well, I tested your branch locally. And I found it works in this way:

```suggestion
            array_shape = Shape(*reversed(shape[:-1])).to_cell_centered()
```

Three position arguments in Shape have default values, so don't need to give them all.

FYI, if you want to type check locally. You should install `mypy` and run `mypy .` under your `yt_idefix` directory.

yt_idefix/data_structures.py
dutta-alankar: Yeah! What's happening with the local test is that I think all the test are not running like the type check when I give `pytest`.

yt_idefix/_io/h5_io.py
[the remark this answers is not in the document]
dutta-alankar: > FYI, if you want to type check locally. You should install `mypy` and run `mypy .` under your `yt_idefix` directory.

Thanks! This was exactly what I was looking for!

dutta-alankar:
@neutrinoceros @xshaokun I have added a tests for xdmf file reading. Please check and let me know. I'm new to pytest and have never written any tests before and have no experience in this. Probably would need help from you guys. I have 9 different kinds of PLUTO outputs (relatively small size, largest being ~15 MB as far as I remember) locally and they are ready to be put in. As suggested by @neutrinoceros , I have not yet committed them yet except for one.

neutrinoceros:
> Tests on the web shows h5py package is missing and hence fails.

To solve this, add the following line to `tests/requirements.txt`
```
h5py>=3.1.0
```

dutta-alankar:
> tests/requirements.txt

Do you mean `requirements/tests.txt`?

yt_idefix/data_structures.py
neutrinoceros: > This is why for now I copied most of the code. This isn't ideal but works out for initial testing.

So everyone is on the same page: as much as I'd like this PR to be "ideal" (minimal diff, no duplicated code ...) I understand the process is easier done progressively, so my only real requirements is that the code works and is tested (we're getting there), and that the code be expressive or at least, minimally commented, to leave room for me to take the wheel and refactor after the fact.

I know the current structure of the code isn't ideal because I never intentionally designed any space for this patch to fill, but if possible I would like to take care of cleaning the rough edges myself rather than having this PR contain large structural changes.

neutrinoceros:
yes, sorry about that

.gitignore
```diff
@@ -2,3 +2,6 @@
 *.egg-info
 **/__pycache__
 dist/*
+venv/*
+tests/data/pluto_disk_cgm_icm/*
+tests/data/pluto_orszag_tang/*
```
neutrinoceros, line 7: these lines do not belong in `.gitignore`. If you mean to ignore these files locally, I suggest putting these lines in `.git/info/exclude` instead (this file is purely local and not meant to be committed)

dutta-alankar: Okay. I'll do that. Do you wish to have all the files that I prepared for testing locally shared in a .zip file? I can readily do that and you can then decide on what to include and what to exclude out. I have around 9 different PLUTO runs with different number of dimensions and geometry combinations.

tests/test_xdmf.py
```diff
@@ -0,0 +1,225 @@
...
+
+def test_class_validation(xdmf_file):
+    file = xdmf_file
+    cls = {
+        "plutoXDMF": PlutoXdmfDataset,
+    }[file["kind"]]
+    assert cls._is_valid(file["path"])
+
+
+def test_parse_pluto_metadata(pluto_xdmf_file):
```
neutrinoceros, line 32: many of these tests appear to be straight up copy/pasted from `test_vtk.py`.

I think copying *all* tests we already have is an overkill, so let's keep just the last 3, starting from `test_slice_plot`, to bring the diff back to a manageable size. I should be able to generalise most of our tests later without duplications.

yt_idefix/_io/commons.py
```diff
@@ -47,6 +47,54 @@ def padded(self) -> Coordinates:
         return Coordinates(arrs[0], arrs[1], arrs[2], self.array_shape)
 
 
+def mapFromCart(
```
neutrinoceros, line 50: avoid `camelCase` in Python, and prefer `snake_case_without_shortened_names`
I assume a better name here would be `get_native_coordinates_from_cartesian` ?

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,102 @@
...
+            nodesX = np.array(nodesX)[0, 0, :]
+            nodesY = np.array(nodesY)[0, :, 0]
+            nodesZ = np.array(nodesZ)[:, 0, 0]
+            array_shape = Shape(*reversed(shape)).to_cell_centered()
+        coords = [nodesX, nodesY, nodesZ]
+    elif geometry in ("polar", "spherical") and dimensions > 1:
+        if dimensions == 2:
+            nodesX = np.array(
+                [
+                    nodesX,
```
neutrinoceros, line 78: a trailing comma causes `black` (via `pre-commit`) to split the code over multiple lines, which in this case hurts readability. Let's drop it (and let black do the formatting again).
```suggestion
                    nodesX
```

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,102 @@
...
+                ]
+            )
+            nodesY = np.array(
+                [
+                    nodesY,
```
neutrinoceros, line 83:
```suggestion
                    nodesY
```

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,102 @@
...
+                ]
+            )
+            nodesZ = np.array(
+                [
+                    nodesZ,
```
neutrinoceros, line 88:
```suggestion
                    nodesZ
```

yt_idefix/_io/commons.py
```diff
@@ -47,6 +47,54 @@ def padded(self) -> Coordinates:
...
+        # on disk coordinates are cell face coordinates.
+        # We need to convert from cartesian to polar,
+        # but no interpolation is needed.
+        r = np.sqrt(xcart[:, 0, 0] ** 2 + ycart[:, 0, 0] ** 2)
+        theta = np.unwrap(np.arctan2(ycart[0, :, 0], xcart[0, :, 0]))
+        z = zcart[0, 0, :]
+
+        coords = [r, theta, z]
+    else:
+        assert geometry == "spherical"
```
neutrinoceros, line 69: Let's be extra careful here since all geometries are not covered yet
```suggestion
    elif geometry == "spherical":
```
... and add the following after this block
```python
    else:
        raise NotImplementedError
```

yt_idefix/api.py
```diff
@@ -5,5 +5,8 @@
     IdefixGrid,
     IdefixVtkDataset,
     IdefixVtkHierarchy,
+    PlutoXdmfHierarchy,
+    PlutoStaticDataset,
```
neutrinoceros, line 9: this reminds me; you should add a line to `pyproject.toml` in the `[project.entry-points."yt.frontends"]` section
In general, you want to make sure you've read *all* places where `PlutoVtkDataset` is listed so you can add an entry for `PlutoXdmfDataset` next to it.

yt_idefix/fields.py
```diff
@@ -20,11 +23,23 @@ def setup_fluid_fields(self):
         setup_magnetic_field_aliases(
             self, "idefix-vtk", [f"BX{idir}" for idir in "123"]
         )
+        for idir in "123":
+            self.alias(
```
neutrinoceros, line 27: why is it needed after `setup_magnetic_field_aliases` ?

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+            self.add_output_field(
+                (self.ds._dataset_type, "temp"),
+                sampling_type="cell",
+                units=unit_system["temperature"],
+            )
+            self.alias(
+                ("gas", "Temperature"),
+                (self.ds._dataset_type, "temp"),
+                units=unit_system["temperature"],
+            )
```
neutrinoceros, line 153: let's avoid non-standard aliases

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+                self.alias(
+                    ("gas", "Tracer %d" % i),
+                    (self.ds._dataset_type, "tr%d" % i),
+                    units="",
+                )
+                self.alias(
+                    ("gas", "tracer %d" % i),
+                    (self.ds._dataset_type, "tr%d" % i),
+                    units="",
+                )
```
neutrinoceros, line 125:
```suggestion
                self.alias(
                    ("gas", f"tracer_{i}"),
                    (self.ds._dataset_type, f"tr{i}"),
                    units="",
                )
```

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+            self.alias(
+                ("gas", "Temperature"),
+                (self.ds._dataset_type, "Temperature"),
+                units=unit_system["temperature"],
+            )
+            self.alias(
+                ("gas", "temperature"),
+                (self.ds._dataset_type, "Temperature"),
+                units=unit_system["temperature"],
+            )
```
neutrinoceros, line 174:
```suggestion
```

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+            self.alias(
+                ("gas", "temperature"),
+                (self.ds._dataset_type, "temp"),
+                units=unit_system["temperature"],
+            )
+        elif (self.ds._dataset_type, "Temperature") in self.field_list:
+            self.add_output_field(
+                (self.ds._dataset_type, "Temperature"),
```
neutrinoceros, line 161:
```suggestion
                (self.ds._dataset_type, "temperature"),
```

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+            self.add_output_field(
+                (self.ds._dataset_type, "temperature"),
+                sampling_type="cell",
+                units=unit_system["temperature"],
+            )
+            self.alias(
+                ("gas", "Temperature"),
+                (self.ds._dataset_type, "temperature"),
+                units=unit_system["temperature"],
+            )
```
neutrinoceros, line 185:
```suggestion
```

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+                sampling_type="cell",
+                function=_mach,
+                units="",
+            )
+            self.alias(
+                ("gas", "mach"),
+                ("gas", "Mach"),
+                units="",
+            )
+
```
neutrinoceros, line 251: the mach number is defined for free by yt if you define the `("gas", "sound_speed")` field here instead

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+            self.alias(
+                ("gas", "number density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
+            self.alias(
+                ("gas", "Number density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
```
neutrinoceros, line 277: please avoid defining multiple aliases per field. Each field should have at most 2 aliases:
`(self.ds._dataset_type, <pluto-standard-name>)` and `("gas", <yt-standard-name>)`

yt_idefix/fields.py
```diff
@@ -47,3 +62,266 @@ def setup_fluid_fields(self):
...
+        self.alias(
+            ("gas", "speed"),
+            ("gas", "Speed"),
+            units=unit_system["velocity"],
+        )
+        self.alias(
+            ("gas", "Velocity Magnitude"),
+            ("gas", "Speed"),
+            units=unit_system["velocity"],
+        )
```
neutrinoceros, line 323: `("gas", "velocity_magnitude")` is already defined by yt itself.
```suggestion
```

yt_idefix/io.py
```diff
@@ -85,3 +87,70 @@ def _read_particle_coords(self, chunks, ptf):
...
+            os.path.basename(self.ds.parameter_filename)
+            .replace(".flt.h5", "")
+            .replace("data.", "")
+            .replace(".dbl.h5", "")
+        )
+        data = {}
+        for field in fields:
+            data[field] = np.empty(size, dtype="float64")
+
+        with h5py.File(self.ds.parameter_filename, "r") as fh:
```
neutrinoceros, line 123:
```suggestion
        with h5py.File(self.ds.filename, "r") as fh:
```

yt_idefix/io.py
```diff
@@ -85,3 +87,70 @@ def _read_particle_coords(self, chunks, ptf):
...
+                            )
+                        elif dimensionality == 2:
+                            field_data = np.array(
+                                [
+                                    field_data,
+                                ]
+                            )
+                        ordering = (2, 1, 0)
+                        # X3 X2 X1 orderding of fields in PLUTO needs to rearranged to X1 X2 X3 order in yt.
+                        values = np.transpose(field_data, ordering)
```
neutrinoceros, line 150:
```suggestion
                        # X3 X2 X1 orderding of fields in PLUTO needs to rearranged to X1 X2 X3 order in yt.
                        values = np.transpose(field_data, axes=(2, 1, 0))
```

.gitignore
neutrinoceros: that's great, feel free to share a link !

tests/conftest.py
```diff
@@ -46,26 +47,39 @@ def load_meta(pdir, meta_file):
...
+            meta_file = pdir / filename
+            if not pdir.is_dir():
+                continue
+            if not meta_file.is_file():
+                continue
+            metadata = load_meta(pdir, meta_file)
+            if "XDMF" in metadata["attrs"]["kind"]:
+                XDMF_FILES.update({metadata["id"]: metadata["attrs"]})
+            else:
+                VTK_FILES.update({metadata["id"]: metadata["attrs"]})
```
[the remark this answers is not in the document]
dutta-alankar: I did this so that multiple `meta` files describing different files in the same directory can be read in for different tests. See for example, `tests/data/pluto_isentropic_vortex`

yt_idefix/_io/commons.py
dutta-alankar: Done!

yt_idefix/io.py
dutta-alankar: Done!

yt_idefix/io.py
dutta-alankar: Done!

yt_idefix/fields.py
dutta-alankar: As far as I can remember, magnetic fields had to be referred to as as for example, ("pluto-vtk", "BX1") and not as conveniently by ("gas", "BX1"). I thought that these lines would perhaps account for that convenience.

yt_idefix/api.py
dutta-alankar: Done!

.gitignore
dutta-alankar: [Here it is](https://indianinstituteofscience-my.sharepoint.com/:u:/g/personal/alankardutta_iisc_ac_in/EXg3BmC-SCpAi6ffgtn1zCwBUcg8g1icENRXEbOtGyD2DA?e=aM8uzi). All of these runs don't have `meta` files. But I can easily add these files to those that you think worth including in the repo.

neutrinoceros: awesome, thank you. I'll give these a look tomorrow !

tests/conftest.py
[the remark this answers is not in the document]
neutrinoceros: Ah, I missed that. Ok then, thanks for the clarification

neutrinoceros:
I haven't had time today to remind myself of how adding large files to this repo works (this is my only project using git-lfs), but I am now having issues checking in and out of this branch locally, seemingly because the couple files you added were committed as ordinary files and shouldn't have.
I will probably need to rewrite the history of this branch sooner than I expected to fix the problem. In the mean time, can you pin point the commit(s) that added `tests/data/pluto_isentropic_vortex/data.0010.vtk` ?

neutrinoceros:
update:
- please install git-lfs locally before committing any data files
- we need to add `*.h5` to the list of file names patterns to be handled with git-lfs (this can be done by running `git las track "*.h5"` and committing `.gitattributes`)
- since mistakes were already made at this point, and a tricky operation might be needed to fix it, I would like to clean up the mess myself, which will involve rewriting the history of your branch again. It is highly unlikely that we can perform this operation without loosing some of your work *unless the branch is perfectly up to date with your local copy at the moment I take the wheel*. So, I'll wait for your explicit permission to do it.

tests/conftest.py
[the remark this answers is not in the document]
dutta-alankar: It is this one: [c667985](https://github.com/dutta-alankar/yt_idefix/commit/c6679a50b81badd997f93a877f2ce6aee6dcf909)
image
Sorry for the late reply. It was night at my time zone and I was sleeping.



> I haven't had time today to remind myself of how adding large files to this repo works (this is my only project using git-lfs), but I am now having issues checking in and out of this branch locally, seemingly because the couple files you added were committed as ordinary files and shouldn't have. I will probably need to rewrite the history of this branch sooner than I expected to fix the problem. In the mean time, can you pin point the commit(s) that added `tests/data/pluto_isentropic_vortex/data.0010.vtk` ?

dutta-alankar:
> I haven't had time today to remind myself of how adding large files to this repo works (this is my only project using git-lfs), but I am now having issues checking in and out of this branch locally, seemingly because the couple files you added were committed as ordinary files and shouldn't have. I will probably need to rewrite the history of this branch sooner than I expected to fix the problem. In the mean time, can you pin point the commit(s) that added `tests/data/pluto_isentropic_vortex/data.0010.vtk` ?

It is this one: [c667985](https://github.com/dutta-alankar/yt_idefix/commit/c6679a50b81badd997f93a877f2ce6aee6dcf909)
image
Sorry for the late reply. It was night at my time zone and I was sleeping.

tests/test_xdmf.py
dutta-alankar: Done!

dutta-alankar:
What would be the way to pass optionally the adiabatic index `gamma` of the gas in `yt.load`? This feature would help accurately determine the sound speed if the simulations are run with a different adiabatic index.

yt_idefix/fields.py
dutta-alankar: The temperature field, these are commonly defined as either "Temp" or "Temperature" by users in PLUTO. So I'll change this accordingly and reduce a lot of these lines.

dutta-alankar:
I think I have been able to now address most of your points suggested regarding changes in `fields.py`.

dutta-alankar:
> update:
> 
> * please install git-lfs locally before committing any data files
> * we need to add `*.h5` to the list of file names patterns to be handled with git-lfs (this can be done by running `git las track "*.h5"` and committing `.gitattributes`) (I'll do this myself, see following poing)
> * since mistakes were already made at this point, and a tricky operation might be needed to fix it, I would like to clean up the mess myself, which will involve rewriting the history of your branch again. It is highly unlikely that we can perform this operation without loosing some of your work _unless the branch is perfectly up to date with your local copy at the moment I take the wheel_. So, I'll wait for your explicit permission to do it.

@neutrinoceros 
Sorry about this! I'm not familiar with `git-lfs` and haven't used it before. I have now followed your suggestion and added `*.h5` data files to `.gitattributes` now to be tracked by `git-lfs`. Let me know if something else can be done about this.

yt_idefix/fields.py
dutta-alankar: Removed these!

yt_idefix/fields.py
dutta-alankar: Noted and changed accordingly!

yt_idefix/_io/h5_io.py
dutta-alankar: Done!

yt_idefix/_io/commons.py
dutta-alankar: Implemented!

yt_idefix/fields.py
dutta-alankar: Noted!

yt_idefix/_io/h5_io.py
dutta-alankar: Fixed!

yt_idefix/fields.py
dutta-alankar: Changed accordingly!

yt_idefix/fields.py
dutta-alankar: Temp and Temperature are most commonly used in the PLUTO community. I have changed this accordingly.

yt_idefix/fields.py
dutta-alankar: Implemented `("gas", "sound_speed")` and removed `mach`. I wasn't aware of this yt feature before.

yt_idefix/fields.py
```diff
@@ -47,3 +62,144 @@ def setup_fluid_fields(self):
...
         )
+        for idir in "123":
+            self.alias(
+                ("gas", f"Vc-BX{idir}"),
+                ("idefix-dmp", f"Vc-BX{idir}"),
+                units="code_magnetic",
+            )
+
+
+class PlutoXdmfFields(FieldInfoContainer):
```
xshaokun, line 73: @dutta-alankar @neutrinoceros The field names of Pluto-vtk and Pluto-xdmf should be the same (not sure about Idefix), so I suggest to use a common class `PlutoField` instead.

dutta-alankar: Good point! However, the names of some of the fields are a bit different in `.vtk` and `.h5` files. The differences are not significant (like `BX1` in `.vtk` is `Bx1` in `.h5`) and perhaps can be merged together.

dutta-alankar:
> What would be the way to pass optionally the adiabatic index `gamma` of the gas in `yt.load`? This feature would help accurately determine the sound speed if the simulations are run with a different adiabatic index.

@xshaokun Any idea how this can be done?

xshaokun:
> What would be the way to pass optionally the adiabatic index `gamma` of the gas in `yt.load`? This feature would help accurately determine the sound speed if the simulations are run with a different adiabatic index.

@dutta-alankar Some frontends have an position argument `parameters` for accepting given parameters, so that users could update `self.gamma` with given value. An example from amrvac frontend:
```
self.gamma = self.parameters.get("gamma", 5.0 / 3.0)
```
Then you could pass your `gamma` by:
```
yt.load("dataset", parameters={"gamma": 4/3})
```

However, I see we have not implement such api in our frontends yet. It's worth to add it!

dutta-alankar:
> > What would be the way to pass optionally the adiabatic index `gamma` of the gas in `yt.load`? This feature would help accurately determine the sound speed if the simulations are run with a different adiabatic index.
> 
> @dutta-alankar Some frontends have an position argument `parameters` for accepting a dict containing given parameters, so that users could update parameters like `self.gamma` with their given value. An example from athena frontend:
> 
> ```
> if "gamma" in self.specified_parameters:
>     self.gamma = float(self.specified_parameters["gamma"])
> else:
>     self.gamma = 5.0 / 3.0
> ```
> 
> Then you could pass your `gamma` by:
> 
> ```
> yt.load("dataset", parameters={"gamma": 4/3})
> ```
> 
> However, I see we have not implemented such an API in our frontends yet. It's worth to add it!
> 
> Remember that `self.parameters` is created in class `Dataset`, so you should update it with the dict passed to option `parameters`.
> 
> If you need some help, I'm willing to fork your repo and implement this API for you :)

Please feel free to fork my repo and add this. I'll merge it then.

neutrinoceros:
> Sorry about this! I'm not familiar with git-lfs and haven't used it before. I have now followed your suggestion and added *.h5 data files to .gitattributes now to be tracked by git-lfs. The only datafiles committed were within 2 MB I think. Let me know if something else can be done about this.

mistakes were made before we had this discussion that cannot be solved additively (i.e., by adding more commits). A rewriting of the branch history is necessary here, and as I said I'd rather to it myself, but won't until you grant me explicit permission. Please make sure to push every changes you have locally before that, as this will be a destructive operation.

yt_idefix/fields.py
neutrinoceros: since field names are slightly different, we cannot fuse these classes. Otherwise, I agree with @xshaokun that we should aim for maximal similarity. This is what yt standardisation is supposed to provide: a common interface that shows little to no difference in usage with respect to the actual data format. For the record this is why I'm advocating against custom field aliases in the `"gas"` field type.

neutrinoceros:
> Some frontends have an position argument parameters for accepting a dict containing given parameters, so that users could update parameters like self.gamma with their given value. An example from athena frontend:

dutta-alankar:
> > Sorry about this! I'm not familiar with git-lfs and haven't used it before. I have now followed your suggestion and added *.h5 data files to .gitattributes now to be tracked by git-lfs. The only datafiles committed were within 2 MB I think. Let me know if something else can be done about this.
> 
> mistakes were made before we had this discussion that cannot be solved additively (i.e., by adding more commits). A rewriting of the branch history is necessary here, and as I said I'd rather to it myself, but won't until you grant me explicit permission. Please make sure to push every changes you have locally before that, as this will be a destructive operation. None of the authors of this PR will be left behind.

Okay. Just let me now when you intend to do this, and I'll push local changes at that time.

dutta-alankar:
> > Some frontends have an position argument parameters for accepting a dict containing given parameters, so that users could update parameters like self.gamma with their given value. An example from athena frontend:
> 
> just a not that adding arguments to `yt.load` (via `MyDataset.__init__`) should be a last resort; if it's possible to parse the parameter value from datafiles that's what we should be doing.

Getting the value of g_gamma is virtually impossible to parse from datafiles.

neutrinoceros:
> Okay. Just let me now when you intend to do this, and I'll push local changes at that time.

I can't know exactly when I'll be able to do this (probably much later today), so it's easier for me if you could give me a 12h window where you know you won't touch this branch. Understand that you'll need my explicit greenlight before you can commit again.

> Getting the value of g_gamma is virtually impossible to parse from datafiles.

To be clear I was using a somewhat broad meaning of "data" here, including `definitions.h` and `pluto.ini`. Maybe this doesn't change your conclusion but just checking everyone is one the same page. If this value can indeed not be retrieved from disk, then I agree we should expose it (add a keyword argument so it can be passed by the user). I would add that it's probably best to never *guess* its value (if it's not provided, renounce providing a default value) so it's much less error prone for users, and more discoverable (it's much easier to realise a field is missing than to understand that values are wrong, and why).

xshaokun:
> I would add that it's probably best to never _guess_ its value (if it's not provided, renounce providing a default value) so it's much less error prone for users, and more discoverable (it's much easier to realise a field is missing than to understand that values are wrong, and why).

I think you convinced me. However, I think it's still safe if these attributes, i.e. `gamma` and `mu`, are initialized only when users give the values. Besides, it opens a window for those who want to leverage more fluid fields instead of define them each time in their scripts. Please review my commit here https://github.com/dutta-alankar/yt_idefix/pull/2 .

dutta-alankar:
> > Okay. Just let me now when you intend to do this, and I'll push local changes at that time.
> 
> I can't know exactly when I'll be able to do this (probably much later today), so it's easier for me if you could give me a 12h window where you know you won't touch this branch. Understand that you'll need my explicit greenlight before you can commit again.
> 
> > Getting the value of g_gamma is virtually impossible to parse from datafiles.
> 
> To be clear I was using a somewhat broad meaning of "data" here, including `definitions.h` and `pluto.ini`. Maybe this doesn't change your conclusion but just checking everyone is one the same page. If this value can indeed not be retrieved from disk, then I agree we should expose it (add a keyword argument so it can be passed by the user). I would add that it's probably best to never _guess_ its value (if it's not provided, renounce providing a default value) so it's much less error prone for users, and more discoverable (it's much easier to realise a field is missing than to understand that values are wrong, and why).

No this value of `g_gamma` defaults to 5/3 in `pluto.h` but this can be changed to something else in `init.c` source file of PLUTO that initializes the simulation.

neutrinoceros:
okay so let's avoid having a default value then

dutta-alankar:
> okay so let's avoid having a default value then

Sure, I'll push this!

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,270 @@ def _get_code_version(self) -> str:
...
+        dataset_type: str | None = None,  # deleguated to child classes
+        units_override: dict[str, UnitLike] | None = None,
+        unit_system: Literal["cgs", "mks", "code"] = "cgs",
+        default_species_fields: Literal["neutral", "ionized"] | None = None,
+        # from here, frontend-specific arguments
+        geometry: Literal["cartesian", "spherical", "cylindrical", "polar"]
+        | None = None,
+        inifile: str | os.PathLike[str] | None = None,
+        definitions_header: str | os.PathLike[str] | None = None,
+        parameters: dict[str, Any] | None = None,
```
neutrinoceros, line 483: I would rather have exposed parameters exposed directly.
```suggestion
        gamma: float | None = None,
        mu: float | None = None,
```

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,270 @@ def _get_code_version(self) -> str:
...
+
+        if "gamma" in self.specified_parameters:
+            self.gamma = float(self.specified_parameters["gamma"])
+        # check g_gamma value in the simulation and pass it. Unfortunately, automating this is not trivial.
+        # For example, > parameters = {'gamma': 5./3.}
+
+        # This is a ballpark number for fully ionized plasma. For more accuracy, say to obtain temperature, let PLUTO dump this field as a user-defined field.
+        if "mu" in self.specified_parameters:
+            self.mu = float(self.specified_parameters["mu"])
+        # For example, > parameters = {'mu': 0.61} must be passed as an argument to yt.load() to plot number density or temperature if these fields are absent in data dump.
```
neutrinoceros, line 505:
```suggestion
        if gamma is not None:
            self.gamma = gamma

        if mu is not None:
            self.mu = mu
```

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,270 @@ def _get_code_version(self) -> str:
...
+
+        # PLUTO (non Chombo) does not support grid refinement
```
neutrinoceros, line 507:
```suggestion
        # PLUTO (non Chombo) does not support AMR
```

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,270 @@ def _get_code_version(self) -> str:
...
+        key = match.group()
+        return str(pluto_def_constants[key])
+
+    def _get_code_version(self) -> str:
+        # take the first line of the header
+        # Code version is read from grid.out
+        """
+        In grid.out
+        # ***********
+        makrs the beginning and the end of header information
```
neutrinoceros, line 615:
```suggestion
        marks the beginning and the end of header information
```

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,270 @@ def _get_code_version(self) -> str:
...
+        with open(self.grid_file) as fh:
+            txt = fh.readlines()
+
+            for _start, line in enumerate(txt):
+                if "# ***********" in line:
+                    count += 1
+                if count == 1:
+                    break
+            _start += 1  # This is the first line of the header info lines in grid.out (version info is here)
+            version = txt[_start].split()[2]
```
neutrinoceros, line 639: Even better would be to match a regular expression, but here's a much simplified version of what you're doing here
```suggestion
        with open(self.grid_file) as fh:
            while not (line := fh.readline()).startswith("# ***********"):
                count += 1
            version = txt[count + 1].split()[2]
```

yt_idefix/fields.py
neutrinoceros: `("gas", "BX1")` should *not* be allowed. It's important to keep a clear distinction between the `"gas"` field type, which should contain fields whose name do not depend on specific conventions used in different codes, and the other, more flexible "pluto-vtk" field type which is explicitly meant to match field names as they are found on disk. 
Please remove this as well as similar blocks in other classes.

yt_idefix/fields.py
```diff
@@ -20,11 +23,23 @@ def setup_fluid_fields(self):
...
 
 class PlutoVtkFields(BaseVtkFields):
     def setup_fluid_fields(self):
         setup_magnetic_field_aliases(self, "pluto-vtk", [f"BX{idir}" for idir in "123"])
+        for idir in "123":
+            self.alias(
+                ("gas", f"BX{idir}"),
+                ("pluto-vtk", f"BX{idir}"),
+                units="code_magnetic",
+            )
```
neutrinoceros, line 42: Following my previous comment
```suggestion
```

neutrinoceros (commented):
another salve of comments. We really need to get the diff smaller now, code duplication is getting a little bit out of control and I currently do not feel confortable with merging such a large change if it means I need to reduce its size by half myself later.

I think we can reasonably set +800 as an upper limit for the diff size (will revise this estimate if needed).

yt_idefix/data_structures.py
dutta-alankar: Okay!

dutta-alankar:
> another salve of comments. We really need to get the diff smaller now, code duplication is getting a little bit out of control and I currently do not feel confortable with merging such a large change if it means I need to reduce its size by half myself later.
> 
> I think we can reasonably set +800 as an upper limit for the diff size (will revise this estimate if needed).

There can be a lot of removal of code in for example `PlutoVtKDataset` which can inherit from `PlutoStaticDataset` now.
Let's point out what would be the places where the code can be reused and I'll try my best to reuse the existing codes making the changes and duplication as minimal as possible.

yt_idefix/data_structures.py
dutta-alankar: Done!

yt_idefix/data_structures.py
dutta-alankar: Done!

yt_idefix/data_structures.py
dutta-alankar: Done!

yt_idefix/fields.py
dutta-alankar: Okay understood.
@neutrinoceros 
One question:
Do we keep `("pluto-vtk", "BX1")` and `("pluto-xdmf", "Bx1")` as seperate or have alias for both as `("PLUTO", "BX1")` say? I think it will be convenient for a user to use the same name for the same field across two different file formats. Let me know your opinion and I'll make changes accordingly.

yt_idefix/data_structures.py
dutta-alankar: I have incorporated your suggestions on this.

yt_idefix/fields.py
xshaokun: > Also the tracer counting logic needs to be put in for `.vtk` files.

@dutta-alankar Could you explain why we need the tracer counting part?

> The differences are not large (like `BX1` in `.vtk` is `Bx1` in `.h5`)

As far as I know, the differences are introduced in this frontend instead of in Pluto (refer to the comment in `vtk_io.read_field_offset_index()`). The variable names of PlutoVtkDataset are uppercased before to avoid duplicating field names. As for now, we should undo this uppercase operation to avoid duplicating Pluto field class. Do you agree? @neutrinoceros

yt_idefix/fields.py
neutrinoceros: moving to a unified name (say `"pluto"`) would technically be a breaking change, so I'm not immediately sold on this.
Additionally, yt supports field type inferring when only the field name is provided (say, `"BX1"`), *on the condition that it matches one and only one field type*, so I encourage using implicit field types for on-disk field names, and explicit field-type, field-name tuples for aliases.

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,260 @@ def _get_code_version(self) -> str:
...
+        return str(pluto_def_constants[key])
+
+    def _get_code_version(self) -> str:
+        # take the first line of the header
+        # Code version is read from grid.out
+        """
+        In grid.out
+        # ***********
+        marks the beginning and the end of header information
+        """
```
neutrinoceros, line 614: I think this comment is now obsolete

yt_idefix/data_structures.py
```diff
@@ -430,6 +458,260 @@ def _get_code_version(self) -> str:
...
+        if not (hasattr(self, "grid_file")):
+            self.grid_file = os.path.join(
+                self.directory, "grid.out"
+            )  # data-loc/grid.out
```
neutrinoceros, line 618: It can be hard to refactor so I'm hesitant to make this comment but `hasattr` should be avoided as much as possible because it makes code much harder to reason about. Its usage may be justified in contexts where maintaining backward compatibility is important, but this is brand new code. There's a lot on your plate already so I can't *demand* that this be refactored out, but please avoid spreading this pattern as I will eventually need to get it of it.

yt_idefix/io.py
```diff
@@ -85,3 +87,59 @@ def _read_particle_coords(self, chunks, ptf):
...
+            1 2.498181e+00 3.500985e-03 747 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            2 4.998045e+00 3.400969e-03 1458 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+            3 7.497932e+00 3.386245e-03 2186 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
+        """
+        entry = int(
+            os.path.basename(self.ds.parameter_filename)
+            .replace(".flt.h5", "")
+            .replace("data.", "")
+            .replace(".dbl.h5", "")
+        )
```
neutrinoceros, line 118:
```suggestion
        if (match := re.search(r"\d{4}", self.ds.basename)) is not None:
            entry = int(match.group())
        else:
            raise RuntimeError(f"Failed to parse entry number from {self.ds.basename}")
```

yt_idefix/io.py
```diff
@@ -85,3 +87,59 @@ def _read_particle_coords(self, chunks, ptf):
...
+        data = {}
+        for field in fields:
+            data[field] = np.empty(size, dtype="float64")
```
neutrinoceros, line 121:
```suggestion
        data = {field: np.empty(size, dtype="float64") for field in fields}
```

yt_idefix/fields.py
```diff
@@ -47,3 +55,137 @@ def setup_fluid_fields(self):
...
+            self.alias(
+                ("gas", "number_density"),
+                (self.ds._dataset_type, "ndens"),
+                units=unit_system["length"] ** -3,
+            )
+
+    def setup_particle_fields(self, ptype):
+        super().setup_particle_fields(ptype)
+        # This will get called for every particle type.
+        # {TODO} Starting with version 4.4 PLUTO has particle support and this function needs an implementation in future
```
neutrinoceros, line 191: Todo comments should start with `TODO:` or `TODO(name):`. I use text editor extensions to discover them so they need to be standardised

yt_idefix/_io/h5_io.py
```diff
@@ -0,0 +1,89 @@
...
+    coords: list[np.ndarray] = []
+    # now assuming that fh is positioned at the end of metadata
+    if geometry in ("cartesian", "cylindrical") or (
+        geometry in ("polar", "spherical") and dimensions == 1
+    ):
+        if dimensions == 1:
+            nodesX = np.array(nodesX)
+            nodesY = np.array(
+                [0.0, 1.0]
+            )  # Default is assumed and not parsed from pluto.ini/grid.out (not present in data)
```
neutrinoceros, line 47: long inline comments tend to hurt readability when formatted with black. The way to work around it is to put the comment on a separate line

neutrinoceros:
I am still waiting for an explicit authorisation on your part to rewrite the branch history. May I do it approximatively 8 hours from now ? (6pm CET)

dutta-alankar:
> I am still waiting for an explicit authorisation on your part to rewrite the branch history. May I do it approximatively 8 hours from now ? (6pm CET)

Yes please do so. Is there something that needs to be done on my part so that you can do so?

neutrinoceros:
You need push any changes you may have locally before I do this, otherwise they'll be lost. You'll also need to manoeuvre back to a usable state after I'm done. I'll provide instructions when I'm done.

dutta-alankar:
> You need push any changes you may have locally before I do this, otherwise they'll be lost. You'll also need to manoeuvre back to a usable state after I'm done. I'll provide instructions when I'm done.

My local repo is upto date with the one on the web.
```
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git pull
Already up to date.
```

yt_idefix/data_structures.py
dutta-alankar: Yes absolutely! Removing this.

yt_idefix/io.py
dutta-alankar: This is much better. Thanks @neutrinoceros !
I had to change `self.ds.basename` to `self.parameter_filename` and it is working.

yt_idefix/io.py
dutta-alankar: Done!

yt_idefix/io.py
neutrinoceros: strange. `self.ds.basename` is strictly equivalent to `os.path.basename(self.ds.parameter_filename)`, so I don't get why you needed to switch back. Also, the `parameter_filename` attribute is an historic alias for the `filename` attribute, please use that one instead.

neutrinoceros: (not only is it strange that `basename` doesn't work, it should be alarming; we certainly do not want to accidentally parse a number from a parent directory's name)

neutrinoceros: (Ah, and another thing: there is an implicit rule that threads shouldn't be resolved by the last person who responded, otherwise we just end up hiding stuff from each other instead of making sure we're all on the same page)

dutta-alankar:
@neutrinoceros 
I have made the final commit incorporating your suggestions before you rewrite the history. I won't be committing anything now unless you have done the history rewrite. Neither will I change my code locally before this.

neutrinoceros:
Excellent. I should be able to do it in a couple hours from now. Will update here when I'm done.

yt_idefix/fields.py
dutta-alankar: I'll correct this after the history rewrite! Sorry I have somehow missed this!

dutta-alankar:
I missed addressing two of your reviews. Should I wait for the rewrite or make a commit right now?

neutrinoceros:
I'm half way through the history-rewriting operation. I may have to take a break (already spent an hour on this. I am not sure I 'll be able to finish this today.

dutta-alankar:
> I'm half way through the history-rewriting operation. I may have to take a break (already spent an hour on this. I am not sure I 'll be able to finish this today.

Sure! No problem at all.

neutrinoceros:
made some progress, but it seems that I've left the code in a weird state where tests don't pass anymore. So far, I do not understand why, and I can't seem to reproduce the problem locally.
@dutta-alankar here's how to get back to a state where you can push to this PR (essentially delete your local copy of the branch and resync with your updated fork)
```
git checkout main
git branch -D xdmf
git fetch origin
git checkout xdmf
```
If this doesn't work immediately and you are unsure what to do, please ping me on slack.
If you want to have a go at fixing CI, feel free to. Otherwise I'll try to fix my mistakes tomorrow. Sorry I couldn't figure it out tonight.

neutrinoceros:
I think I need to rewrite the commit in which I add the data again, but it seems I need additional authorisations for that. @dutta-alankar, could you invite me to collaborate on your fork ?

dutta-alankar:
> I think I need to rewrite the commit in which I add the data again, but it seems I need additional authorisations for that. @dutta-alankar, could you invite me to collaborate on your fork ?

Yes sure. Can you please tell me how to do that?

dutta-alankar:
I think I was able to do it. Can you please confirm?

neutrinoceros:
yup, you did correctly, thank you !
It's still not allowing me to push the correct version of the data-adding commit, so I think I'll just to it as a separate PR before I loose my mind

neutrinoceros:
I'm working on it in #219, but the current issue seems to be that we went over my (free) git-lfs quota, so I'm currently unable to push new data to the repo (same reason it didn't work when I tried to push to yours).
I'll continue trying to fix it, hopefully I won't need to do anything too dramatic.

dutta-alankar:
Please let me know if I can do anything to contribute and help.

neutrinoceros:
I will. I'll try to resolve the git-lfs shenanigans over the weekend, for now I got CI green again on this branch (albeit with no actual tests done with XDMF data)

neutrinoceros:
I'm going to do some small refactors here to leverage #220

neutrinoceros:
@dutta-alankar since it's currently impossible to push test data I will be running tests locally for now. If it's okay with you I'd like to continue working on your branch and start deduplicating code myself.

dutta-alankar:
> @dutta-alankar since it's currently impossible to push test data I will be running tests locally for now. If it's okay with you I'd like to continue working on your branch and start deduplicating code myself.

Yes sure. Please go ahead!

neutrinoceros:
> As long as all authors are in contribution list this won't be a problem.

This won't be an issue, I've co-credited everyone on the amalgamated commit :)

I made a bunch of small PRs in support to this one (mostly arrangements to the testing infrastructure), and I'm now able to test locally (since we still cannot commit the new datasets). I will now be focusing my efforts on deduplication.

neutrinoceros:
I've made a lot of progress in deduplicating the code. There's probably still room to further simplify the code, though I'm going to stop here for today. Please take a moment to review the current state of the code (I may have made some agressive changes here and there), and report if this is anything still missing (apart from test data).

@xshaokun, I'd like your opinion on my solution regarding equation of state parameters (`gamma` and `mu`). At the moment it is still only available for XDMF datasets, though it should be easy to move the code up to the parent class so PlutoVtkDataset can benefit from it.

xshaokun:
> @xshaokun, I'd like your opinion on my solution regarding equation of state parameters (`gamma` and `mu`).

@neutrinoceros Thanks for asking, I'm only wondering that, why not initialize them as attributes of dataset so that they can be called in yt's `setup_fluid_fields` and we don't need to define some fields like `_sound_speed` by ourself.

neutrinoceros:
Good point ! indeed we get sound speed for free if we define gamma as an attribute, I missed that (TBH I don't like this design). However `mu` doesn't bring anything automatically, and I would like to keep the API similar for both options, so I'm hesitant to rely on the way yt currently generates the sound speed field.

I should point that the one field we get for free by having `ds.gamma` defined is 3 lines of code, so it seems reasonable to duplicate to avoid relying on it, now that I'm tempted to remove it in yt itself 😄 
```python
    def _sound_speed(field, data):
        tr = data.ds.gamma * data[ftype, "pressure"] / data[ftype, "density"]
        return np.sqrt(tr)
```

neutrinoceros:
> While trying to clone and checkout, I get some messages related to test dataset now. Please have a look at this. Are these pointers related messages expected and normal?

@dutta-alankar this is git-lfs complaining that I made mistakes when using it (that happened too far back for me to go back and rewrite the main branch's history). you can fix it with
```
git lfs uninstall
git reset --hard
git lfs install
git lfs pull
```

xshaokun:
> Good point ! indeed we get sound speed for free if we define gamma as an attribute, I missed that (TBH I don't like this design). However `mu` doesn't bring anything automatically, and I would like to keep the API similar for both options, so I'm hesitant to rely on the way yt currently generates the sound speed field.

That makes sense!

> However `mu` doesn't bring anything automatically

Just reminding, `mu` can be calculated for free in yt only for the simplest cases (fully ionized or neutral without metal) by a builtin function `compute_mu`. In other cases, `mu` is required to be given to calculate number density :

        def _number_density(field, data):
            mu = getattr(data.ds, "mu", compute_mu(data.ds.default_species_fields))
            return data[ftype, "density"] / (pc.mh * mu) 

Otherwise users have to override it, although this field is defined shortly as well :)

So far, the simplest cases are enough for me,  so I've never dig into it. We can consider this problem when we actually need it!

neutrinoceros:
Wow, impressive ! How did you even find this ? it didn't cross my mind to grep for `getattr` because it's an overkill in this context: fields in yt may be constructed with various implicit requirements, and they won't crash any application when these requirements are not met. Anyway, good catch !

> So far, the simplest cases are enough for me, so I've never dig into it. We can consider this problem when we actually need it!

that would be my preference too ! @dutta-alankar, do you actually need this now, or would it be okay to take this out of your PR and focus on more fundamental stuff for the time being ?

dutta-alankar:
> Wow, impressive ! How did you even find this ? it didn't cross my mind to grep for `getattr` because it's an overkill in this context: fields in yt may be constructed with various implicit requirements, and they won't crash any application when these requirements are not met. Anyway, good catch !
> 
> > So far, the simplest cases are enough for me, so I've never dig into it. We can consider this problem when we actually need it!
> 
> that would be my preference too ! @dutta-alankar, do you actually need this now, or would it be okay to take this out of your PR and focus on more fundamental stuff for the time being ?

@neutrinoceros You mean the availibility of `mu` to calculate number density? This is optional and maybe skipped. Anyways, one can always provide a `ndens` field that can be dumped.

neutrinoceros:
> You mean the availibility of mu to calculate number density? 

yes. Let's drop this part for now then.

neutrinoceros:
oh and because it's always nicer to be explicit: you can back control over you PR, I won't be pushing anything else for now :)

dutta-alankar:
> oh and because it's always nicer to be explicit: you can back control over you PR, I won't be pushing anything else for now :)

I'm also a little bit confused. Can you please tell me if someone wishes to pass `mu` and `gamma` explicitly, is there any way to do so using `yt.load`?

neutrinoceros:
`yt.load` accepts arbitrary keyword arguments, but the responsibility of using them belongs to `Dataset` subclasses, so if we want that, we need to support it here. I have nothing against doing that, but it doesn't have to be done in *this* PR, or does it ? The smaller the patches, the easier it'll be to review and revisit later :)

dutta-alankar:
> `yt.load` accepts arbitrary keyword arguments, but the responsibility of using them belongs to `Dataset` subclasses, so if we want that, we need to support it here. I have nothing against doing that, but it doesn't have to be done in _this_ PR, or does it ? The smaller the patches, the easier it'll be to review and revisit later :)

I got it! Let's make a separate PR for that if needed.

neutrinoceros:
Cool ! This means we can remove everything related to mu and gamma here

dutta-alankar:
> Cool ! This means we can remove everything related to mu and gamma here

I'll do that and make a commit.

dutta-alankar:
@neutrinoceros 
I see the files required for testing shows up in git status. Should I go ahead with the commit?
```
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git status
On branch xdmf
Your branch is up to date with 'origin/xdmf'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   tests/data/cataclysmic-disk/data.0086.vtk
        modified:   tests/data/khi/data.0100.vtk
        modified:   tests/data/khi/dump.0100.dmp
        modified:   tests/data/pluto_disk_planet/data.0010.vtk
        modified:   tests/data/pluto_rotor/data.0030.vtk
        modified:   tests/data/pluto_sedov/data.0018.vtk
        modified:   tests/data/pluto_sod/data.0001.vtk
        modified:   yt_idefix/data_structures.py
        modified:   yt_idefix/fields.py

no changes added to commit (use "git add" and/or "git commit -a")
```

neutrinoceros:
These are not staged, so committing is safe. I think `git stash` might help to get rid of these "modifications". Any way, right now GitHub will reject any push yt_idefix (even forks) that include git-lfs stuff, so you should be *unable* to mess things up at the moment :)

dutta-alankar:
> These are not staged, so committing is safe. I think `git stash` might help to get rid of these "modifications". Any way, right now GitHub will reject any push yt_idefix (even forks) that include git-lfs stuff, so you should be _unable_ to mess things up at the moment :)

Great! But now I'm unable to push further commits. Can you help now?
```
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git stash
Saved working directory and index state WIP on xdmf: 888c378 ENH: Added PLUTO xdmf dump support
Encountered 7 file(s) that should have been pointers, but weren't:
        tests/data/cataclysmic-disk/data.0086.vtk
        tests/data/khi/data.0100.vtk
        tests/data/khi/dump.0100.dmp
        tests/data/pluto_disk_planet/data.0010.vtk
        tests/data/pluto_rotor/data.0030.vtk
        tests/data/pluto_sedov/data.0018.vtk
        tests/data/pluto_sod/data.0001.vtk
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git status
On branch xdmf
Your branch is up to date with 'origin/xdmf'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   tests/data/cataclysmic-disk/data.0086.vtk
        modified:   tests/data/khi/data.0100.vtk
        modified:   tests/data/khi/dump.0100.dmp
        modified:   tests/data/pluto_disk_planet/data.0010.vtk
        modified:   tests/data/pluto_rotor/data.0030.vtk
        modified:   tests/data/pluto_sedov/data.0018.vtk
        modified:   tests/data/pluto_sod/data.0001.vtk

no changes added to commit (use "git add" and/or "git commit -a")
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git add .
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git commit -m "Deferred mu and gamma defaults to a future PR"
[xdmf 547bcbb] Deferred mu and gamma defaults to a future PR
 7 files changed, 0 insertions(+), 0 deletions(-)
 rewrite tests/data/cataclysmic-disk/data.0086.vtk (99%)
 rewrite tests/data/khi/data.0100.vtk (99%)
 rewrite tests/data/khi/dump.0100.dmp (99%)
 rewrite tests/data/pluto_disk_planet/data.0010.vtk (99%)
 rewrite tests/data/pluto_rotor/data.0030.vtk (99%)
 rewrite tests/data/pluto_sedov/data.0018.vtk (100%)
 rewrite tests/data/pluto_sod/data.0001.vtk (100%)
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git push
Username for 'https://github.com': ^C
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ pre-commit run black --all-files
black....................................................................Passed
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git push
Username for 'https://github.com': dutta-alankar
Password for 'https://dutta-alankar@github.com':
Username for 'https://github.com': dutta-alankar
Password for 'https://dutta-alankar@github.com':
batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.
error: failed to push some refs to 'https://github.com/dutta-alankar/yt_idefix.git'
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ git stash
No local changes to save
```

neutrinoceros:
You staged the test files with `git add .`. Let's undo the latest commit (keeping the changes as unstaged)
```
git reset HEAD^
```
now you can add files you need to commit *explicitly*, commit, and push.
I recommend using `git commit -p -m "message"` to make "partial" commits more easily.

Don't use `git add .` in general, it will get you into trouble by staging any files you didn't mean to commit that wasn't already listed by `.gitignore`.

neutrinoceros:
I don't know why your copy keeps trying to "modify" test files, supposedly this is fixed by running the script I showed earlier: https://github.com/neutrinoceros/yt_idefix/pull/209#issuecomment-1455682262

did you try it ?

dutta-alankar:
> I don't know why your copy keeps trying to "modify" test files, supposedly this is fixed by running the script I showed earlier: [#209 (comment)](https://github.com/neutrinoceros/yt_idefix/pull/209#issuecomment-1455682262)
> 
> did you try it ?

Yes I did. Let me delete everything and try to re-clone.

dutta-alankar:
@neutrinoceros Can you help? I'm unable to push commits.

neutrinoceros:
While this state of affairs is annoying, there's no error in the log you showed me. To me it still looks like you should be able to push _as long as you're careful not to include datafiles in your commits_. You can make partial commits with `git commit -p ...` or just manually stage files you *do* want to commit with `git add myfile1 myfile2 && git commit ...`

dutta-alankar:
> While this state of affairs is annoying, there's no error in the log you showed me. To me it still looks like you should be able to push _as long as you're careful not to include datafiles in your commits_. You can make partial commits with `git commit -p ...` or just manually stage files you _do_ want to commit with `git add myfile1 myfile2 && git commit ...`

Thanks! Instead of `git add .` I had to individually add the python files and then pushing the commit was successful.

dutta-alankar:
I got a `pre-commit` failure. 
`pre-commit run black --all-files` shows passed in my local copy. @neutrinoceros Can you help?

yt_idefix/data_structures.py
```diff
@@ -759,3 +788,110 @@ def _is_valid(cls, filename: str, *args, **kwargs) -> bool:
...
+
+    def __init__(
+        self,
+        filename,
+        *,
+        eos_params: dict[str, float] | None = None,
+        **kwargs,
+    ):
+        super().__init__(filename, **kwargs)
+        self.parameters["equation_of_state"] = eos_params or {}
```
neutrinoceros, line 806: @dutta-alankar  this part should be removed too

dutta-alankar: Let's keep the `__init__()` function but remove the `equation_of_state` line. This will allow easier incorporation of this or some other features in future by passing to `yt.load()`

yt_idefix/data_structures.py
```diff
@@ -759,3 +788,109 @@ def _is_valid(cls, filename: str, *args, **kwargs) -> bool:
...
+    _field_info_class = PlutoXdmfFields
+
+    def __init__(
+        self,
+        filename,
+        *,
+        eos_params: dict[str, float] | None = None,
+        **kwargs,
+    ):
+        super().__init__(filename, **kwargs)
```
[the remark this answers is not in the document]
dutta-alankar: Yes sure. That I understand. What I'm suggesting is we can retain this basic `__init__()` structure here so that we can add any new feature for later and the path size will be smaller. If you don't want this then I can remove this redundant part as well.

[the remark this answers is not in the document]
neutrinoceros: In general I don't want to add code unless it's immediately useful. 
Here's it's actually worse than useless because it introduces an unused keyword argument, so I'd like this removed for now.

dutta-alankar:
> FYI I deactivated pre-commit.ci auto fixing last week, so you need to perform fixes locally or to trigger them manually. I recommend installing pre-commit locally.

But I get nothing with `pre-commit`
```
(venv) (test) alankar@juggernaut:~/Documents/myrepos/yt_idefix$ pre-commit run black --all-files
black....................................................................Passed
```

neutrinoceros:
black isn't the one complaining (ruff is). Run all hooks with

```
pre-commit run --all-files
```

neutrinoceros:
I'm now able to test this locally with multiple geometries and so far it seems good. I'd like to perform a couple more minor adjustments, is it okay with you if I take back the wheel one last time in this PR ? I should be able to get it across the finish line tomorrow and cut a release.

dutta-alankar:
> I'm now able to test this locally with multiple geometries and so far it seems good. I'd like to perform a couple more minor adjustments, is it okay with you if I take back the wheel one last time in this PR ? I should be able to get it across the finish line tomorrow and cut a release.

Sure. Please go ahead. You may also try the data that I shared a few days ago. It has different combinations of geometry and dimensions.

neutrinoceros:
That's the data I'm using ! Thank you

neutrinoceros:
I revised the class hierarchy and discovered that `PlutoXdmfHierarchy` is broken. Currently it's missing an implementation for `_cell_width` and `_cell_centers`, so it cannot be instantiated. It used to get away with it by inheriting these methods from `IdefixVtkHierarchy` (now renamed `VtkHierarchy`), but they were never actually functional.

@dutta-alankar can you implement them ? In this case I would prefer having an implementation that's 99% copy-pasted than using inheritance. I won't be pushing again to this branch for now.

dutta-alankar:
> I revised the class hierarchy and discovered that `PlutoXdmfHierarchy` is broken. Currently it's missing an implementation for `_cell_width` and `_cell_centers`, so it cannot be instantiated. It used to get away with it by inheriting these methods from `IdefixVtkHierarchy` (now renamed `VtkHierarchy`), but they were never actually functional.
> 
> 
> 
> @dutta-alankar can you implement them ? In this case I would prefer having an implementation that's 99% copy-pasted than using inheritance. I won't be pushing again to this branch for now.

Sure I can do that. I initially inherited to prevent code duplication.

dutta-alankar:
I'm off to Switzerland for a conference this week and may not be able to do this entirely. But I'll try to wrap this up as much as I can this week itself.

neutrinoceros:
Yes

dutta-alankar:
@neutrinoceros I had to do a few changes and now this seems to be working again. Please have a look.

neutrinoceros:
Thanks, I should be able to take a look tomorrow.

neutrinoceros:
Tests are passing again, so I think this is now good to go. I should be able to merge and publish this later today, thanks again for your patience and dedication !

dutta-alankar:
> Tests are passing again, so I think this is now good to go. I should be able to merge and publish this later today, thanks again for your patience and dedication !

Thanks a lot for helping me to finish this.

neutrinoceros:
pushing a final history rewrite just to have a single atomic commit on the branch, and then I'll publish

dutta-alankar:
Regarding the test data for `flt.h5` and `bl.h5`, what to do make it available with the repo?

neutrinoceros:
We can't do anything until April, my (free) bandwidth limit has been reached for this month. I'll give it another go in time, and for now I'll just run tests locally if need be.

PR #219, opened 2023-03-03
TST: add XDMF test data

neutrinoceros:


neutrinoceros:
I am currently unable to push relevant `*.vtk` and `.h5` files due to my git-lfs quota being overloaded.

neutrinoceros:
So the issue I'm having is that I went by and large over my bandwidth quota for this month. As a result I am unable to push any new data for next 23 days ! I guess we will have to compose with this restriction for now.
For information, ping @dutta-alankar

tests/conftest.py
```diff
@@ -45,26 +46,39 @@ def load_meta(pdir, meta_file):
...
-    meta_file = pdir / "meta.yaml"
-    if not pdir.is_dir():
-        continue
-    if not meta_file.is_file():
-        continue
-
-    metadata = load_meta(pdir, meta_file)
-    VTK_FILES.update({metadata["id"]: metadata["attrs"]})
+    for filename in os.listdir(pdir):
+        if filename.startswith("meta") and filename.endswith(".yaml"):
```
neutrinoceros, line 50: I took this straight out of #209 but this system looks like it can be improved to limit metadata duplication. I think it would be reasonable to replace the `"path"` scalar with a list (`"paths"`)

tests/data/pluto_isentropic_vortex/meta00.yaml
```diff
@@ -0,0 +1,16 @@
+id: orszag_tang
+attrs:
+  path: data.0010.vtk
+  has_geometry: true
+  geometry: cartesian
+  dimensionality: 2
+  kind: plutoVTK
```
neutrinoceros, line 7: the fact that we need to distinguish two varieties of pluto files indicate that we should probably have more attributes (`code` and `format` seem like reasonable names)

neutrinoceros: handled in #221

tests/conftest.py
neutrinoceros: handled in #221

neutrinoceros:
tests are currently failing with a `FileNotFoundError` (hidden behind a warning...), which is expected, since I cannot yet push the actual datasets.

neutrinoceros:
pre-commit.ci autofix

PR #245, opened 2023-03-30
BUG: add warning and error messages for missing inifile

### LINKED ISSUE #244: Error encountered while parsing `definitions.h` on trying to make a plot (2D cartesian geometry)
I' attaching the python script and the data file that generates this error here as a [GDrive link](https://drive.google.com/file/d/12IIA_LOFwZhD_MHA-M6Q-21applEQFpD/view?usp=sharing).

```
Traceback (most recent call last):
  File "/home/alankar/Documents/myrepos/data-pluto/tests/lecoanet/test.py", line 4, in <module>
    ds = yt.load("./data.%04d.dbl.h5"%60)
  File "/home/alankar/Documents/myrepos/yt_idefix/venv/lib/python3.9/site-packages/yt/loaders.py", line 107, in load
    return candidates[0](fn, *args, **kwargs)
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 342, in __init__
    super().__init__(
  File "/home/alankar/Documents/myrepos/yt_idefix/venv/lib/python3.9/site-packages/yt/data_objects/static_output.py", line 261, in __init__
    self._parse_parameter_file()
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 857, in _parse_parameter_file
    super()._parse_parameter_file()
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 519, in _parse_parameter_file
    super()._parse_parameter_file()
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 388, in _parse_parameter_file
    self._parse_definitions_header()
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 652, in _parse_definitions_header
    expr = re.sub(r"g_inputParam\[(\w+)\]", self._get_input_parameter, expr)
  File "/home/alankar/anaconda3/envs/test/lib/python3.9/re.py", line 210, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "/home/alankar/Documents/myrepos/yt_idefix/yt_idefix/data_structures.py", line 662, in _get_input_parameter
    return str(self.parameters["Parameters"][key])
KeyError: 'Parameters'
```

xshaokun:
This PR fixes #244 

- [x] Add proper messages

- [x] Add a test for missing inifile

### DIFF
--- a/tests/data/pluto_disk_planet/incorrect_inifile
+++ b/tests/data/pluto_disk_planet/incorrect_inifile
@@ -1,3 +1,7 @@
+# Compared with the original pluto.ini,
+# this inifile lacks the required parameters
+# under section [Parameters] for unit definitions
+
 [Grid]
 X1-grid    1  0.4          128  u  2.5
 X2-grid    1  1.448623279  16   u  1.570796327
--- a/tests/data/pluto_disk_planet/incorrect_inifile
+++ b/tests/data/pluto_disk_planet/incorrect_inifile
@@ -1,0 +1,48 @@
+[Grid]
+X1-grid    1  0.4          128  u  2.5
+X2-grid    1  1.448623279  16   u  1.570796327
+X3-grid    1  0.0          384  u  6.283185307
+
+[Chombo Refinement]
+Levels             4
+Ref_ratio          2     2  2  2  2
+Regrid_interval    2     2  2  2
+Refine_thresh      0.3
+Tag_buffer_size    3
+Block_factor       8
+Max_grid_size      64
+Fill_ratio         0.75
+
+[Time]
+CFL            0.25
+CFL_max_var    1.1
+tstop          1.0
+first_dt       1.e-4
+
+[Solver]
+Solver    hllc
+
+[Boundary]
+X1-beg    userdef
+X1-end    userdef
+X2-beg    reflective
+X2-end    eqtsymmetric
+X3-beg    periodic
+X3-end    periodic
+
+[Static Grid Output]
+uservar     0
+dbl         -1.0  -200  single_file
+flt         -1.0  -1    single_file
+vtk         0.1   -1    single_file
+tab         -1.0  -1
+ppm         -1.0  -1
+png         -1.0  -1
+log         1
+analysis    -1.0  100
+
+[Chombo HDF5 output]
+Checkpoint_interval    -1.0  0
+Plot_interval          1.0   0
+
+[Parameters]
--- a/tests/data/pluto_disk_planet/meta.yaml
+++ b/tests/data/pluto_disk_planet/meta.yaml
@@ -9,7 +9,6 @@ attrs:
   dimensionality: 3
   kind: pluto
   has_units: true
-  require_inifile: true
   units:
     length: 7.77909038e13 cm
     velocity: 208457.8557 cm/s
--- a/tests/data/pluto_disk_planet/meta.yaml
+++ b/tests/data/pluto_disk_planet/meta.yaml
@@ -9,6 +9,7 @@ attrs:
   dimensionality: 3
   kind: pluto
   has_units: true
+  require_inifile: true
   units:
     length: 7.77909038e13 cm
     velocity: 208457.8557 cm/s
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -1,6 +1,7 @@
 import os
 import re
 import shutil
+from pathlib import Path
 
 import pytest
 from more_itertools import distinct_combinations
@@ -99,26 +100,23 @@ def test_pluto_two_units_override(vtk_file_with_units):
     assert_allclose_units(ds.mass_unit, expect_mass)
 
 
-def test_missing_inifile(vtk_file_with_units, tmp_path):
-    file = vtk_file_with_units
-    if file.get("require_inifile"):
-        tmpdir = tmp_path / "missing_inifile"
-        shutil.copytree(
-            file["path"].parent, tmpdir, ignore=shutil.ignore_patterns("*.ini")
-        )
-        with pytest.warns(
-            UserWarning, match=r"The inifile is missing for unit definitions. *"
-        ):
-            yt.load(list(tmpdir.glob("*.vtk"))[0])
+def test_missing_inifile(tmp_path):
+    tmpdir = tmp_path / "missing_inifile"
+    shutil.copytree(
+        Path(__file__).parent / "data/pluto_disk_planet",
+        tmpdir,
+        ignore=shutil.ignore_patterns("*.ini"),
+    )
+    with pytest.warns(
+        UserWarning, match=r"The inifile is missing for unit definitions. *"
+    ):
+        yt.load(tmpdir / "data.0010.vtk")
 
 
-def test_incorrect_inifile(vtk_file_with_units):
-    file = vtk_file_with_units
-    if file.get("require_inifile"):
-        inifile = file["path"].parent / "incorrect_inifile"
-        if inifile.exists:
-            with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
-                yt.load(file["path"], inifile=inifile)
+def test_incorrect_inifile():
+    file_dir = Path(__file__).parent / "data/pluto_disk_planet"
+    with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
+        yt.load(file_dir / "data.0010.vtk", inifile=file_dir / "incorrect_inifile")
 
 
 def test_pluto_invalid_units_override(pluto_vtk_file):
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -103,7 +103,7 @@ def test_pluto_two_units_override(vtk_file_with_units):
 def test_missing_inifile(tmp_path):
     tmpdir = tmp_path / "missing_inifile"
     shutil.copytree(
-        Path(__file__).parent / "data/pluto_disk_planet",
+        Path(__file__).parent / "data" / "pluto_disk_planet",
         tmpdir,
         ignore=shutil.ignore_patterns("*.ini"),
     )
@@ -114,7 +114,7 @@ def test_missing_inifile(tmp_path):
 
 
 def test_incorrect_inifile():
-    file_dir = Path(__file__).parent / "data/pluto_disk_planet"
+    file_dir = Path(__file__).parent / "data" / "pluto_disk_planet"
     with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
         yt.load(file_dir / "data.0010.vtk", inifile=file_dir / "incorrect_inifile")
 
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -1,5 +1,6 @@
 import os
 import re
+import shutil
 
 import pytest
 from more_itertools import distinct_combinations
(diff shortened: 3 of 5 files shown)


xshaokun:
This PR gives a warning message for all cases where the inifile is missing, and only raises `RuntimeError` for the place where the inifile is explicitly needed. 

So far, inifile is only required for unit definition. For our data samples, it's only required by `pluto_disk_planet`

yt_idefix/data_structures.py
```diff
@@ -370,6 +370,9 @@ def _get_meta_file(
         ):
             return str(file.absolute())
         else:
+            ytLogger.warning(
```
neutrinoceros, line 373: I'd rather issue an actual warning instead of a logger warning.

xshaokun:
Just changed the function name for the test :)

neutrinoceros (commented):
I see two competing approaches here:
- raise a warning and declare that behaviour without an inifile is undefined (an error might be raised, but maybe it won't)
- unconditionally raise an error (in which case we don't need a warning)

I think in this case it's reasonable to raise a warning *and* try to avoid crashing (units might be wrong, but if we're loading with `units_system='code'` we might not care).

yt_idefix/data_structures.py
xshaokun: Could you explain what's the difference between them and how to choose them?

xshaokun: Thanks, I got it!

xshaokun:
I'm not sure about whether  such warning message is clear enough to stop users, especially those are new to yt, from doing wrong things. 

The test should be refactored because I found it cannot mock missing inifile.

neutrinoceros:
Do you mean you need help to refactor the test ? I'm happy to give you a hand, but I'd rather not review too early in your process.

xshaokun:
Oh I mean that I need to modify the test in this PR, because I thought `inifile=None` can mock missing inifile but it can't. I just found a fixture named `tmp_path` in the document of pytest, which might be useful here. Let me try it!

neutrinoceros:
yeah, you could mock the desired situation using a combination of `tmp_path` and [`os.symlink`](https://docs.python.org/3/library/os.html#os.symlink)

xshaokun:
I used `shutil.copytree` here instead because it's easier to remove inifile. And `tmp_path` will be removed after 3 sessions by default, so we don't need to concern about the disk usage.

xshaokun:
Do you think we need `test_incorrect_inifile`? Maybe I should remove it. I guess the corresponding warning will only be triggered by users instead of development in the future.

yt_idefix/data_structures.py
```diff
@@ -658,7 +658,20 @@ def _parse_definitions_header(self) -> None:
...
+            warnings.warn(
+                "The inifile is missing for unit definitions. "
+                f"Specify it to the keyword argument 'inifile' (default: {self._default_inifile}), "
+                "or make sure that all fields are in code units!"
+            )
+            return "1.0"
         key = match.group(1)
+        if key not in self.parameters["Parameters"]:
+            warnings.warn(
+                f"Cannot get the value of {key} from {self._inifile}."
```
neutrinoceros, line 671: this is the only time we actually require `self._inifile` in this function, so it would make more sense to check if it exists _now_ and adapt the warning message as needed

dutta-alankar:
I wanted to suggest something related to this and let you decide if you want to have this feature. How about passing custom locations of 'definitions.h' or 'pluto.ini' in yt.load() and having the default location to be the same as the data files like what it is now?

neutrinoceros:
@dutta-alankar this is exactly what the `inifile` and `definitions_header` arguments are for
https://github.com/neutrinoceros/yt_idefix/blob/f971592d0bc099b502c1b130e6e7ea6b4e93857c/yt_idefix/data_structures.py#L328-L329
If there's any problem (or if you have questions) with how they work, feel free to open an issue !

tests/data/pluto_disk_planet/incorrect_inifile
```diff
@@ -0,0 +1,48 @@
+[Grid]
```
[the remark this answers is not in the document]
xshaokun: I removed the parameters that are needed for unit definitions in this case.

yt_idefix/data_structures.py
xshaokun: I just considered a case where there is only an incorrect inifile in the directory, maybe the `pluto.ini` generated by Pluto's setup script. In this case, although the inifile exists, the value of the specific parameter still cannot be returned and it will raise an error.

It might be a very rare case. I just want to make it clear where the error was raised. If you think it's unnecessary, I can drop it and the corresponding test.

tests/data/pluto_disk_planet/incorrect_inifile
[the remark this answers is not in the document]
neutrinoceros: It would help if you could include this explanation as a comment in the file itself then

yt_idefix/data_structures.py
xshaokun: @neutrinoceros I guess you might miss this message :) Could you give me your opinion on my consideration? Then I will modify it according to your requests.

A brief summary to help you recall the work in this PR: I designed two warning messages for the cases where inifile is missing and an incorrect inifile is given (without required parameters inside), and also two corresponding tests.

neutrinoceros: Thanks for reminding me, indeed it seems I missed your first reply !
I'm not arguing *against* the new warning you're adding. What I meant is that the earlier check (currently at line 661) should be moved here, because this function shouldn't *rely on* `self._inifile`: it's just convenient to include it in the current warning, but really we don't need it. Is that any clearer ?

xshaokun: Sure, I get your point now!

tests/data/pluto_disk_planet/meta.yaml
```diff
@@ -9,6 +9,7 @@ attrs:
   dimensionality: 3
   kind: pluto
   has_units: true
+  require_inifile: true
```
neutrinoceros, line 12: It doesn't seem worth adding a one-time key to the test infrastructure

tests/test_vtk.py
```diff
@@ -98,6 +99,28 @@ def test_pluto_two_units_override(vtk_file_with_units):
...
+            yt.load(list(tmpdir.glob("*.vtk"))[0])
+
+
+def test_incorrect_inifile(vtk_file_with_units):
+    file = vtk_file_with_units
+    if file.get("require_inifile"):
+        inifile = file["path"].parent / "incorrect_inifile"
+        if inifile.exists:
+            with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
+                yt.load(file["path"], inifile=inifile)
```
neutrinoceros, line 121: I'd much rather have these tests not use fixtures and parametrization since they really are meant to run on a single test case

neutrinoceros (commented):
I think we can still simplify the tests. Otherwise, looking good !

tests/test_vtk.py
xshaokun: Okay, I'll explicitly pass the only path of sample data.

tests/test_vtk.py
```diff
@@ -99,26 +100,23 @@ def test_pluto_two_units_override(vtk_file_with_units):
...
-            file["path"].parent, tmpdir, ignore=shutil.ignore_patterns("*.ini")
-        )
-        with pytest.warns(
-            UserWarning, match=r"The inifile is missing for unit definitions. *"
-        ):
-            yt.load(list(tmpdir.glob("*.vtk"))[0])
+def test_missing_inifile(tmp_path):
+    tmpdir = tmp_path / "missing_inifile"
+    shutil.copytree(
+        Path(__file__).parent / "data/pluto_disk_planet",
```
neutrinoceros, line 106: for windows compat
```suggestion
        Path(__file__).parent / "data" / "pluto_disk_planet",
```

tests/test_vtk.py
```diff
@@ -99,26 +100,23 @@ def test_pluto_two_units_override(vtk_file_with_units):
...
 
-def test_incorrect_inifile(vtk_file_with_units):
-    file = vtk_file_with_units
-    if file.get("require_inifile"):
-        inifile = file["path"].parent / "incorrect_inifile"
-        if inifile.exists:
-            with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
-                yt.load(file["path"], inifile=inifile)
+def test_incorrect_inifile():
+    file_dir = Path(__file__).parent / "data/pluto_disk_planet"
```
neutrinoceros, line 117: Same as above
```suggestion
    file_dir = Path(__file__).parent / "data" / "pluto_disk_planet"
```

xshaokun:
I just rebase this PR to the main branch, so that you can just squash it.

tests/test_vtk.py
```diff
@@ -98,6 +100,25 @@ def test_pluto_two_units_override(vtk_file_with_units):
...
+    )
+    with pytest.warns(
+        UserWarning, match=r"The inifile is missing for unit definitions. *"
+    ):
+        yt.load(tmpdir / "data.0010.vtk")
+
+
+def test_incorrect_inifile():
+    file_dir = Path(__file__).parent / "data" / "pluto_disk_planet"
+    with pytest.warns(UserWarning, match=r"Cannot get the value of *"):
```
neutrinoceros, line 118: Clarification: the `match` argument expects a regexp pattern. In particular this means
- `' *'` is interpreted as "any number of `' '`characters" (I assume you meant "anything next should match")
- matching a string only means that the *start* of the string fits the patterns, so we don't actually need to fill the whole message or do anything special to express that the rest of the string can be ignored.
```suggestion
    with pytest.warns(UserWarning, match=r"Cannot get the value of"):
```

tests/test_vtk.py
```diff
@@ -98,6 +100,25 @@ def test_pluto_two_units_override(vtk_file_with_units):
...
 
+def test_missing_inifile(tmp_path):
+    tmpdir = tmp_path / "missing_inifile"
+    shutil.copytree(
+        Path(__file__).parent / "data" / "pluto_disk_planet",
+        tmpdir,
+        ignore=shutil.ignore_patterns("*.ini"),
+    )
+    with pytest.warns(
+        UserWarning, match=r"The inifile is missing for unit definitions. *"
```
neutrinoceros, line 111:
```suggestion
        UserWarning, match=r"The inifile is missing for unit definitions."
```

tests/test_vtk.py
xshaokun: Oh I thought it would pass only when the whole warning message is matched.

> I assume you meant "anything next should match"

That's true, I thought it is like that in shell.

neutrinoceros (approved):

neutrinoceros:
Thanks again @xshaokun !
I want to iterate on warning messages again before I release this patch, but I didn't want to hold your PR, so I'll open another one soon and tag you as a reviewer if that's okay with you !

xshaokun:
> I want to iterate on warning messages again before I release this patch, but I didn't want to hold your PR, so I'll open another one soon and tag you as a reviewer if that's okay with you !

No problem! I really need some help on that :)

PR #251, opened 2023-04-18
BUG: do not normalise field names to upper case in Idefix vtk

vdbma:
Idefix allows the user to output custom fields in the vtk files and their name is chosen by the user. This means that names containing lower case character may be chosen by the user. Moreover the default vector potential field name also contains a lower case character (_e.g._ `AX1e`)
Currently, for normalisation reasons with different Pluto versions, all native field names are converted to upper case.
This means that such fields can't be accessed via _e.g._ `("idefix-vtk","AX1e")`.

This PR proposes to use this normalisation for Pluto datasets only. This allows the fields to be accessed with the same name they have in the vtk file.

### DIFF
--- a/yt_idefix/_io/vtk_io.py
+++ b/yt_idefix/_io/vtk_io.py
@@ -204,7 +204,9 @@ def read_grid_coordinates(
     return Coordinates(coords[0], coords[1], coords[2], array_shape)
 
 
-def read_field_offset_index(fh: BinaryIO, shape: Shape, upper_case_varnames:bool) -> dict[str, int]:
+def read_field_offset_index(
+    fh: BinaryIO, shape: Shape, upper_case_varnames: bool
+) -> dict[str, int]:
     # assuming fh is correctly positioned (read_grid_coordinates must be called first)
     retv: dict[str, int] = {}
 
--- a/yt_idefix/_io/vtk_io.py
+++ b/yt_idefix/_io/vtk_io.py
@@ -205,7 +205,7 @@ def read_grid_coordinates(
 
 
 def read_field_offset_index(
-    fh: BinaryIO, shape: Shape, upper_case_varnames: bool
+    fh: BinaryIO, shape: Shape, *, upper_case_varnames: bool
 ) -> dict[str, int]:
     # assuming fh is correctly positioned (read_grid_coordinates must be called first)
     retv: dict[str, int] = {}
--- a/yt_idefix/_io/vtk_io.py
+++ b/yt_idefix/_io/vtk_io.py
@@ -204,7 +204,7 @@ def read_grid_coordinates(
     return Coordinates(coords[0], coords[1], coords[2], array_shape)
 
 
-def read_field_offset_index(fh: BinaryIO, shape: Shape) -> dict[str, int]:
+def read_field_offset_index(fh: BinaryIO, shape: Shape, upper_case_varnames:bool) -> dict[str, int]:
     # assuming fh is correctly positioned (read_grid_coordinates must be called first)
     retv: dict[str, int] = {}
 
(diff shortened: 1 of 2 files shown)


yt_idefix/data_structures.py
```diff
@@ -725,11 +725,22 @@ def _parse_parameter_file(self):
...
         # from here self.geometry is assumed to be set
 
+        if self.dataset_type == "idefix-vtk":
+            # idefix may have upper and lower case
+            # characters in the variable names
+            upper_case_varnames = False
+        else:
+            # For Pluto it depends on the version
+            # for normalisation purposes they are
+            # we use upper case names
```
neutrinoceros, line 735:
```suggestion
            # we normalize to upper case
```

yt_idefix/data_structures.py
```diff
@@ -725,11 +725,22 @@ def _parse_parameter_file(self):
...
+            upper_case_varnames = True
+
         # parse the grid
         with open(self.filename, "rb") as fh:
             coords = vtk_io.read_grid_coordinates(fh, geometry=self.geometry)
+            "ld"
```
neutrinoceros, line 741: looks like some unused left over
```suggestion
```

yt_idefix/_io/vtk_io.py
```diff
@@ -204,7 +204,9 @@ def read_grid_coordinates(
     return Coordinates(coords[0], coords[1], coords[2], array_shape)
 
 
-def read_field_offset_index(fh: BinaryIO, shape: Shape) -> dict[str, int]:
+def read_field_offset_index(
+    fh: BinaryIO, shape: Shape, upper_case_varnames: bool
```
neutrinoceros, line 208: Let's add a default value for simplicity
```suggestion
    fh: BinaryIO, shape: Shape, upper_case_varnames: bool = False
```

yt_idefix/data_structures.py
```diff
@@ -725,11 +725,22 @@ def _parse_parameter_file(self):
...
 
+        if self.dataset_type == "idefix-vtk":
+            # idefix may have upper and lower case
+            # characters in the variable names
+            upper_case_varnames = False
+        else:
+            # For Pluto it depends on the version
+            # for normalisation purposes they are
+            # we use upper case names
+            upper_case_varnames = True
```
neutrinoceros, line 736:
```suggestion
        # some versions of Pluto define field names in lower case
        # so we normalize to upper case to avoid duplicating data
        # in IdefixVtkFieldInfo.known_other_fields
        normalize_varnames = self.dataset_type == "pluto-vtk"
```

yt_idefix/_io/vtk_io.py
```diff
@@ -218,7 +220,8 @@ def read_field_offset_index(fh: BinaryIO, shape: Shape) -> dict[str, int]:
         # some versions of Pluto define field names in lower case
         # so we normalize to upper case to avoid duplicating data
         # in IdefixVtkFieldInfo.known_other_fields
```
neutrinoceros, line 222: let's drop this comment (move it to where we set this parameter)

yt_idefix/data_structures.py
```diff
@@ -725,11 +725,22 @@ def _parse_parameter_file(self):
...
+
         # parse the grid
         with open(self.filename, "rb") as fh:
             coords = vtk_io.read_grid_coordinates(fh, geometry=self.geometry)
+            "ld"
             self._field_offset_index = vtk_io.read_field_offset_index(
-                fh, coords.array_shape
+                fh, coords.array_shape, upper_case_varnames
```
neutrinoceros, line 743:
```suggestion
                fh, coords.array_shape, upper_case_varnames=normalize_varnames
```

vdbma:
Thanks for the review. It should be in better shape now.

PR #254, opened 2023-04-20
UX: rephrase a warning

neutrinoceros:
follow up to #245

I would also like to do something about the second warning because it's a problem that it's not possible to silence it.
I'm thinking we should probably just raise a hard error is receiving an invalid `pluto.ini` file, but is it actually invalid to leave units out of it ?

ping @xshaokun for review

### DIFF
--- a/yt_idefix/data_structures.py
+++ b/yt_idefix/data_structures.py
@@ -662,11 +662,13 @@ class StaticPlutoDataset(GoodboyDataset, ABC):
         """Replace matched input parameters with its value"""
         key = match.group(1)
         if key not in self.parameters.get("Parameters", {}):
-            if not self._inifile:
+            if not os.path.exists(self._inifile):
                 warnings.warn(
-                    "The inifile is missing for unit definitions. "
-                    f"Specify it to the keyword argument 'inifile' (default: {self._default_inifile}), "
-                    "or make sure that all fields are in code units!",
+                    f"Could not find inifile ({self._inifile}). "
+                    "Inferred code units might be inaccurate\n"
+                    "To silence this warning, specify the inifile keyword argument to yt.load, "
+                    "as a path (either absolute or relative to the data file's parent directory). "
+                    f"Default is {self._default_inifile!r}. ",
                     stacklevel=2,
                 )
             else:
--- a/tests/test_vtk.py
+++ b/tests/test_vtk.py
@@ -107,9 +107,7 @@ def test_missing_inifile(tmp_path):
         tmpdir,
         ignore=shutil.ignore_patterns("*.ini"),
     )
-    with pytest.warns(
-        UserWarning, match=r"The inifile is missing for unit definitions."
-    ):
+    with pytest.warns(UserWarning, match="Could not find inifile"):
         yt.load(tmpdir / "data.0010.vtk")
 
 


yt_idefix/data_structures.py
```diff
@@ -662,11 +662,13 @@ def _get_input_parameter(self, match: re.Match) -> str:
...
-            if not self._inifile:
+            if not os.path.exists(self._inifile):
                 warnings.warn(
-                    "The inifile is missing for unit definitions. "
-                    f"Specify it to the keyword argument 'inifile' (default: {self._default_inifile}), "
-                    "or make sure that all fields are in code units!",
+                    f"Could not find inifile ({self._inifile}). "
+                    "Inferred code units might be inaccurate\n"
+                    "To silence this warning, specify the inifile keyword argument to yt.load, "
+                    "as a path relative to the data file's parent directory "
```
xshaokun, line 670: I think an absolute path is also acceptable here, right? Because `self._get_meta_file()` is able to accept an absolute path.

xshaokun:
> I'm thinking we should probably just raise a hard error is receiving an invalid `pluto.ini` file, but is it actually invalid to leave units out of it ?

Do you mean skipping to initialize units when the inifile is missing? I think it cannot be skipped. That's the part implemented in yt's `Dataset.__init__()` that we cannot touch in this repo. That's the reason I have to return some value when the parameters are invalid in `StaticPlutoDataset_get_input_parameter()`.

And as you said:

> I think in this case it's reasonable to raise a warning and try to avoid crashing (units might be wrong, but if we're loading with units_system='code' we might not care).

I'm totally agreed with that ! I think it's better to just prompt an warning, as long as it can accurately inform the hidden danger.

neutrinoceros:
ah right, thanks for putting up with my contradictions !
So indeed, keeping the warning is the way forward. Should it be silenced in case `unit_system="code"` was passed ?

yt_idefix/data_structures.py
neutrinoceros: good catch, I forgot about that !

neutrinoceros: (updated)

xshaokun:
> Should it be silenced in case `unit_system="code"` was passed ?

I'm not certain about that. Since somewhat goes wrong, maybe we should keep the warning here until the bug is actually fixed. 

However, this warning would be prompted several times if there are more than one input parameter being used in unit definitions. I missed this point. We can let it warn just once :)

neutrinoceros:
>  maybe we should keep the warning until the bug is actually fixed.

what bug ?

> this warning would be prompted several times if there are more than one input parameter being used in unit definitions

that is incorrect. By default, the Python interpreter raises warnings *once per location*, see https://docs.python.org/3/library/warnings.html#the-warnings-filter

neutrinoceros:
> Ah I mean missing the inifile.

I think we're talking about two different things. I was referring to the other warning, which I haven't touched in this PR yet:
https://github.com/neutrinoceros/yt_idefix/blob/f5e87be6678371cddd80e4b7cb3b5ad9c21c6fff/yt_idefix/data_structures.py#L673-L677

xshaokun:
> I think we're talking about two different things. I was referring to the other warning, which I haven't touched in this PR yet:

I was also talking about this and don't have a definite opinion on that, because someone can still convert the fields into cgs units manually, via `.in_cgs()`. While I think users are unlikely to intend to do it when they set `unit_system = "code"`.

neutrinoceros:
Ok then let's just roll with it for now, it's a rare enough case that it's probably not worth spending too much time discussing it until it actually becomes a problem for some one. Does this seem fair to you ?

xshaokun:
> Does this seem fair to you ?

Sure!

PR #255, opened 2023-04-22
RFC: reorganize field classes

### LINKED ISSUE #253: Merge the field class of Pluto-vtk and Pluto-xdmf
Here I would like to discuss the feasibility of using a common field class for both Pluto-vtk and Pluto-xdmf.

So far, Pluto-vtk and Idefix-vtk share a common vtk field class. To do this, the upper case is applied to Pluto-vtk's field name.

However, I guess the field names should be more common for Pluto-vtk and Pluto-xdmf, and the field class should be _classified by code instead of output format_. By doing this, maybe we can drop the upper case operation on Pluto-vtk's field name and it would be easier to maintain the field class.

Actually, I started using Pluto in version 4.4, so I don't know about the field name in the version before that. What about your opinion? @neutrinoceros @dutta-alankar

xshaokun:
This PR will close #253 

The test `test_derived_field(vtk_file)` should also be redesigned because the field names are different between Pluto-vtk and Idefix-vtk. Do you have any suggestion on that? @neutrinoceros


neutrinoceros:
> The test test_derived_field(vtk_file) should also be redesigned because the field names are different between Pluto-vtk and Idefix-vtk

It looks like the reason it's breaking is that you removed case normalisation completely, was this intended ? I'd rather we don't do that because it's a breaking change.

xshaokun:
> It looks like the reason it's breaking is that you removed case normalisation completely, was this intended ?

Yes, I have to make the case consistent before merging two fields. If we have to keep the case normalization in order to compatible with older version of Pluto, I think it should be applied to Pluto-xdmf as well.

neutrinoceros:
> If we have to keep the case normalization in order to compatible with older version of Pluto

My concern is more about keeping compatibility with existing scripts using `yt-idefix`.

> I think it should be applied to Pluto-xdmf as well.

that would also be a breaking change (normalising to upper case), but given that support for Pluto-xdmf is relatively new, I think it's still negotiable, if @dutta-alankar agrees to it.

neutrinoceros:
additional thoughts: if we're going to normalise case for all Pluto fields, we should normalise to whatever current versions of Pluto uses, which I guess would be lower case ?

xshaokun:
> My concern is more about keeping compatibility with existing scripts using `yt-idefix`.

I think your concern is reasonable but it also means we have to refactor the field classes sooner, otherwise it will be costly if we do it in the future or develop two Pluto field classes separately. 

> we should normalise to whatever current versions of Pluto uses, which I guess would be lower case ?

Field names of Pluto in output are not case normalized, referring to those in `PlutoXdmfFields`. But in the code, they are defined as macros in upper case, for example in MHD:
```c
#define  RHO 0

#define  MX1 1
#define  MX2 2
#define  MX3 3
#define  BX1 4
#define  BX2 5
#define  BX3 6

#if HAVE_ENERGY
  #define ENG  7
  #define PRS  ENG
#endif
#if DIVB_CONTROL == DIV_CLEANING
  #define PSI_GLM  (7 + HAVE_ENERGY)
#endif

#define VX1   MX1
#define VX2   MX2
#define VX3   MX3
``` 

If we're going to normalize case for all Pluto fields, upper case might be more consistent for coding.

dutta-alankar:
> > If we have to keep the case normalization in order to compatible with older version of Pluto
> 
> My concern is more about keeping compatibility with existing scripts using `yt-idefix`.
> 
> > I think it should be applied to Pluto-xdmf as well.
> 
> that would also be a breaking change (normalising to upper case), but given that support for Pluto-xdmf is relatively new, I think it's still negotiable, if @dutta-alankar agrees to it.

I'm okay with changing `pluto-xdmf`. I'll be busy with other work for the next few weeks and won't be able to dedicate much time to this but I support the cleaner code layout proposed here.

dutta-alankar:
> > My concern is more about keeping compatibility with existing scripts using `yt-idefix`.
> 
> I think your concern is reasonable but it also means we have to refactor the field classes sooner, otherwise it will be costly if we do it in the future or develop two Pluto field classes separately.
> 
> > we should normalise to whatever current versions of Pluto uses, which I guess would be lower case ?
> 
> Field names of Pluto in output are not case normalized, referring to those in `PlutoXdmfFields`. But in the code, they are defined as macros in upper case, for example in MHD:
> 
> ```c
> #define  RHO 0
> 
> #define  MX1 1
> #define  MX2 2
> #define  MX3 3
> #define  BX1 4
> #define  BX2 5
> #define  BX3 6
> 
> #if HAVE_ENERGY
>   #define ENG  7
>   #define PRS  ENG
> #endif
> #if DIVB_CONTROL == DIV_CLEANING
>   #define PSI_GLM  (7 + HAVE_ENERGY)
> #endif
> 
> #define VX1   MX1
> #define VX2   MX2
> #define VX3   MX3
> ```
> 
> If we're going to normalize case for all Pluto fields, upper case might be more consistent for coding.

Here is Pluto 4.2: [https://gitlab.mpcdf.mpg.de/sdoetsch/PLUTO-4.2](https://gitlab.mpcdf.mpg.de/sdoetsch/PLUTO-4.2)
`var_names.c` contain the names of PLUTO fields that get stored during IO. Inside the code, the fields are referred to by macros and accessed from a 4D array.

xshaokun:
> Here is Pluto 4.2: https://gitlab.mpcdf.mpg.de/sdoetsch/PLUTO-4.2

Many thanks!

xshaokun:
I recognize the builtin fields by the field names listed in `.known_other_fields`. Currently they're only the part of fields which are able to be output. Shall I complete the list?

yt_idefix/fields.py
```diff
@@ -51,16 +49,52 @@ def setup_fluid_fields(self):
...
+        ("EX1S", ("code_magnetic/c", ["EX1s"], None)),
+        ("EX2S", ("code_magnetic/c", ["EX2s"], None)),
+        ("EX3S", ("code_magnetic/c", ["EX3s"], None)),
+        ("CRG", ("", ["code_magnetic/code_length/c"], None)),
+        ("PSI_GLM", ("", [], None)),
+        ("PHI_GLM", ("", [], None)),
+        ("RHO_D", ("code_density", [], None)),
+        ("VX1_D", ("code_velocity", [], None)),
+        ("VX2_D", ("code_velocity", [], None)),
+        ("VX3_D", ("code_velocity", [], None)),
```
neutrinoceros, line 84: out of curiosity, does the `_D` suffix mean "dust" ?

xshaokun: yes

yt_idefix/fields.py
```diff
@@ -54,84 +47,63 @@ def setup_fluid_fields(self):
...
+        ("AX1", ("code_magnetic", [], None)),
+        ("AX2", ("code_magnetic", [], None)),
+        ("AX3", ("code_magnetic", [], None)),
+        ("BX1", ("code_magnetic", [], None)),
+        ("BX2", ("code_magnetic", [], None)),
+        ("BX3", ("code_magnetic", [], None)),
+        ("EX1", ("code_magnetic/c", [], None)),
+        ("EX2", ("code_magnetic/c", [], None)),
+        ("EX3", ("code_magnetic/c", [], None)),
+        ("BX1S", ("code_magnetic", ["BX1s"], None)),
```
neutrinoceros, line 72: Why the alias (`BX1S` to `BX1s`) ?

yt_idefix/_io/vtk_io.py
```diff
@@ -217,7 +219,10 @@ def read_field_offset_index(
         s = line.decode()
         datatype, varname, dtype = s.split()
 
-        if upper_case_varnames:
+        # some versions of Pluto define field names in lower case
+        # so we normalize builtin field (not complete yet) to upper case
```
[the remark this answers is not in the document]
xshaokun: Got it!

yt_idefix/fields.py
xshaokun: `BX1s` is the macro defined in Pluto for variable `Bx1`, so are others (for staggered MHD). 

I set Pluto macro names as alias if they're different to upper case of the output names, so that users can call the fields with the names they use in Pluto.

xshaokun: I found it makes unconsistency that some fields can be called with ("gas", ` pluto-macro-name`) while some fields can only be called with (`datatype`, `pluto-macro-name`). So I think `pluto-macro-name` should be set as alias for all of them, or let those macro names go instead. Which one do you prefer?

xshaokun:
I have no idea why the test fails online. They all pass locally, with python3.10 on ubuntu and python3.8 on mac. Need some help here. @neutrinoceros

neutrinoceros:
I don't know if I'll be able to look into it today but I'll try my best.

xshaokun:
> I don't know if I'll be able to look into it today but I'll try my best.

No hurry :)

neutrinoceros:
Oh actually I think you just hit a flaky test that _sometimes_ crashes with a runtime warning. I haven't had time to investigate it yet but it should go away after a retry or two.

yt_idefix/fields.py
xshaokun: ping @neutrinoceros again for a response (I guess my replies were missed :)

yt_idefix/fields.py
```diff
@@ -51,9 +51,9 @@ class PlutoFields(BaseFields):
     known_other_fields = (
         # PLUTO 4 standard variable names normalized in upper case,
         # Referring to Tools/IDL/pload.pro in Pluto v4.2-patch2
-        # Ion fractions are not listed here
+        # Field names of ion fractions are excluded to normalize
         # Each entry here is of the form
-        # ( "name", ("units", ["fields", "to", "alias"], # "display_name")),
```
neutrinoceros, line 66: feel free to rephrase this comment, but it should still reflect that more than one alias can be set

yt_idefix/fields.py
```diff
@@ -54,84 +47,63 @@ def setup_fluid_fields(self):
...
-    _erg_units = "code_mass * (code_length/code_time)**2"
-    _rho_units = "code_mass / code_length**3"
-    _mom_units = "code_mass / code_length**2 / code_time"
-    _vel_units = "code_length/code_time"
-
+class PlutoFields(BaseFields):
     known_other_fields = (
+        # PLUTO 4 standard variable names normalized in upper case,
+        # Referring to Tools/IDL/pload.pro in Pluto v4.2-patch2
+        # Field names of ion fractions are excluded to normalize
```
neutrinoceros, line 54:
```suggestion
        # Field names of ion fractions are not normalized to upper case
```

xshaokun:
Rebase to the main branch and add the last two commits

yt_idefix/io.py
```diff
@@ -118,8 +118,10 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
                 for grid in chunk.objs:
                     nd = 0
                     for field in fields:
-                        ftype, fname = field
-                        position = f"/Timestep_{entry}/vars/{fname}"
+                        _, fname = field
+                        position = (
+                            f"/Timestep_{entry}/vars/{self.ds._field_name_map[fname]}"
+                        )
```
neutrinoceros, line 124: Unless black is forcing this style (I don't think it does)
```suggestion
                        position = f"/Timestep_{entry}/vars/{self.ds._field_name_map[fname]}"
```

yt_idefix/fields.py
neutrinoceros: indeed, sorry I often seem to miss stuff these days...
yes, these aliases, quite confusingly, are used to map to the `"gas"` field type, so it seems preferable to not have them at all.

yt_idefix/io.py
neutrinoceros: alright, not critical

yt_idefix/fields.py
xshaokun: > so it seems preferable to not have them at all.

Ok I'll do it.

yt_idefix/fields.py
neutrinoceros: do you happen to know wether these represent a pressure less fluid or if it's the result of particle deposition ?

xshaokun: I just checked it. I think they represent a pressure less fluid, which is introduced in a related file `dust_fluid.c` but not written in the user guide.

xshaokun: BTW, there is another type of dust as a kind of particle mentioned in the user guide. This module is still under construction and not public.

neutrinoceros: Good to know ! Thank you very much.
I'll make a last review later today and probably merge + release

yt_idefix/_io/vtk_io.py
```diff
@@ -217,7 +219,10 @@ def read_field_offset_index(
         s = line.decode()
         datatype, varname, dtype = s.split()
 
-        if upper_case_varnames:
+        # some versions of Pluto define field names in lower case
+        # so we normalize standard output field names to upper case
+        # to avoid duplicating data in PlutoFields.known_other_fields
+        if varname.upper() in dict(field_info.known_other_fields).keys():
```
neutrinoceros, line 225: just noticed this. I'd prefer we use `more_itertools` for these kinds of loops and avoid conversions.
One possible implementation would be
```suggestion
        if varname.upper() in unzip(field_info.known_other_fields)[0]:
```
(requires`from more_itertools import unzip`)
`more_itertools` is currently not listed as a hard dependency to `yt_idefix` (but it *is* already a dependency in `yt` itself), so it would need to be added to `project.dependencies` in `pyproject.toml`.

It also strikes me that this list is all that we need from the `field_info` argument, so maybe it should be the argument instead ?

yt_idefix/data_structures.py
```diff
@@ -885,6 +870,26 @@ def _parse_parameter_file(self):
...
+            root = list(h5f.keys())[0]
+            self._detected_field_list = list(h5f[f"{root}/vars/"])
+            self._field_name_map = {}
+            # some versions of Pluto define field names in lower case
+            # so we normalize standard output field names to upper case
+            # to avoid duplicating data in PlutoFields.known_other_fields
+            for varname in self._detected_field_list:
+                if (
+                    varname.upper()
+                    in dict(self._field_info_class.known_other_fields).keys()
```
neutrinoceros, line 883: same as above.
To add to my previous comment, iterating over a dictionary itself is the same as iterating over its keys, so `.keys()` is redundant here

neutrinoceros (approved):
a couple more nits actually

yt_idefix/_io/vtk_io.py
xshaokun: Thanks! First time to know it.

> It also strikes me that this list is all that we need from the `field_info` argument, so maybe it should be the argument instead ?

It sounds reasonable. I'll change it.

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "inifix>=4.1.0",
     "numpy>=1.17.3",
     "yt>=4.1.0",
+    "more-itertools>=5.0",
```
neutrinoceros, line 38: good job finding the minimal version here !

yt_idefix/_io/vtk_io.py
```diff
@@ -205,7 +205,7 @@ def read_grid_coordinates(
 
 
 def read_field_offset_index(
-    fh: BinaryIO, shape: Shape, *, upper_case_varnames: bool
+    fh: BinaryIO, shape: Shape, *, default_field_list: map
```
neutrinoceros, line 208: Thanks to this type hint, I see that my last comments mislead you, let's correct this.
A `map` is a generator (a one-time-only iterator), and `in` doesn't work as you'd think:
```python
>>> it = map(lambda x : 1, range(5))
>>> 1 in it
False
>>> list(it)
[]
```

Because we need to iterate over this more than once within the `while` loop, we need to convert it to a list beforehand
```suggestion
    fh: BinaryIO, shape: Shape, *, default_field_list: list[str]
```

yt_idefix/data_structures.py
```diff
@@ -742,16 +735,13 @@ def _parse_parameter_file(self):
...
-        normalize_varnames = self.dataset_type == "pluto-vtk"
-
         # parse the grid
         with open(self.filename, "rb") as fh:
             coords = vtk_io.read_grid_coordinates(fh, geometry=self.geometry)
             self._field_offset_index = vtk_io.read_field_offset_index(
-                fh, coords.array_shape, upper_case_varnames=normalize_varnames
+                fh,
+                coords.array_shape,
+                default_field_list=unzip(self._field_info_class.known_other_fields)[0],
```
neutrinoceros, line 744:
```suggestion
                default_field_list=list(unzip(self._field_info_class.known_other_fields)[0]),
```

yt_idefix/data_structures.py
```diff
@@ -885,6 +873,26 @@ def _parse_parameter_file(self):
...
+            root = list(h5f.keys())[0]
+            self._detected_field_list = list(h5f[f"{root}/vars/"])
+            self._field_name_map = {}
+            # some versions of Pluto define field names in lower case
+            # so we normalize standard output field names to upper case
+            # to avoid duplicating data in PlutoFields.known_other_fields
+            for varname in self._detected_field_list:
+                if (
+                    varname.upper()
+                    in unzip(self._field_info_class.known_other_fields)[0]
```
neutrinoceros, line 886:
```suggestion
                    in list(unzip(self._field_info_class.known_other_fields)[0])
```

neutrinoceros (commented):
I see I've misguided you in my last round of reviews, sorry about that.

yt_idefix/_io/vtk_io.py
xshaokun: Ah I have tested it by
```python
>>> field, info = unzip(ds.field_info.known_other_fields)
>>> "RHO" in field
True
```
Then I thought it should work. I should try to run it again :)

yt_idefix/data_structures.py
xshaokun: I guess a list comprehension can also be used here and is more readable, right?
```suggestion
                default_field_list=[f[0] for f in self._field_info_class.known_other_fields],
```

neutrinoceros: that's a good point, and it's actually much more readable. Thanks !

yt_idefix/data_structures.py
```diff
@@ -881,10 +882,9 @@ def _parse_parameter_file(self):
             # so we normalize standard output field names to upper case
             # to avoid duplicating data in PlutoFields.known_other_fields
             for varname in self._detected_field_list:
-                if (
-                    varname.upper()
-                    in unzip(self._field_info_class.known_other_fields)[0]
-                ):
+                if varname.upper() in [
+                    f[0] for f in self._field_info_class.known_other_fields
+                ]:
```
neutrinoceros, line 887: we should still store the result of the list comp rather than recomputing it for every iteration of the parent for loop

xshaokun: Got it！

neutrinoceros (approved):

neutrinoceros:
Thank you again for the time you put in this @xshaokun ! I've said this already, but I'm really glad to have you on board with this project !

neutrinoceros:
I changed the title of the PR because, as a technically breaking change (we intentionally removed some aliases), it doesn't fully qualify as a "refactor", and I want to highlight that in the release notes.

xshaokun:
> Thank you again for the time you put in this @xshaokun ! I've said this already, but I'm really glad to have you on board with this project !

I've also learned a lot in this process. l'm also grateful for your patient reviews!

PR #276, opened 2023-06-21
[DOC] Minor correction in documentation

dutta-alankar:
Minor correction of quote type in documentation.


neutrinoceros (approved):
nice catch, thank you !

PR #325, opened 2024-08-31
TST: bump uv (0.2.25 -> 0.4.1)

neutrinoceros:
Newer versions log which dependencies were built, which is particularily useful in bleeding-edge CI


neutrinoceros:
Hum, looks like I was wrong... maybe `Built yt` is only a transient message that happens to sometime persist on my system but doesn't necessarily on CI. In any case this doesn't hurt, so merging any way.

PR #336, opened 2024-10-03
Added the ability to read xdmf dumps from `Idefix`

dutta-alankar:
@neutrinoceros As discussed a few days ago, here is the first implementation of a working reader to parse Idefix xdmf dumps. 

While developing the HDF5 IO module on `Idefix`, I have taken a special attention to make the files contain additional attributes that make them self-contained and hence don't need `definitions.hpp` or `idefix.ini` to process the data. I am aware that my implementation is far from being the cleanest, and simply does the job for the time being. But I believe that it is a good head start and can be trimmed and matured to a polished version. I'm open to take in comments to improve this.

To test, I'm using the following code with a KHI data dump from the following [PR](https://github.com/idefix-code/idefix/pull/270). I'm also sharing the data used in this test here in this [link](https://drive.google.com/file/d/15wrzLV6i_SS9vXWiGiobRzEhVNl7xGmJ/view?usp=sharing).

```
import yt
import yt_idefix

num = 1
ds = yt.load(f"data.{num:04d}.flt.h5")

# Create gas density slices in all three axes.
p = yt.SlicePlot(ds, "z", ("gas", "density") # , width=(4.0, "cm")
             )
p.set_cmap(field=("gas", "density"), cmap="viridis")
p.save()
```

which gives the following output:

data 0001 dbl h5_Slice_z_density

I compared this with the following ParaView output:

data 0001 dbl h5_Slice_z_density_Paraview

which seems to agree well with each other.


neutrinoceros:
This is awesome, thanks a lot for doing this !
I will make sure to review in detail soon, but just a couple very quick comments right off the bat:

In you test script, note that 
```python
import yt_idefix
```
can be omitted if you have `yt>=4.2.0` (which I think `yt_idefix` requires anyway).
In the upcoming yt 4.4.0, you'll also be able to install as `pip install 'yt[idefix]'` so users don't even need to know that yt_idefix is a separate package.

Otherwise, regarding the PR itself, keep in mind we'll need to add at least one formal test for it, but as it'll require using `git-lfs` (which I'm allergic to), we can do that part later / separately.

neutrinoceros:
quick update: I have a lot going on at the moment and I don't think I'll be able to prioritize this review before next week, but I'm not forgetting about it. Thanks for your patience !

dutta-alankar:
No hurries at all.

neutrinoceros:
I just pushed a couple changes to the class hierarchy and xdmf tests so that parametrized tests still pass with your dataset included (but I didn't commit the data itself yet).

neutrinoceros (changes requested):
Sorry this took me so long to get back to you. I pushed a couple simplifications myself and added some additional comments and suggestions. I think we need to try and deduplicate as much as possible before I can merge.

src/yt_idefix/data_structures.py
```diff
@@ -928,6 +899,210 @@ def _parse_parameter_file(self):
...
+        self.parameters["definitions"] = {}
+        self.parameters["definitions"]["geometry"] = self.attributes["geometry"]
+
+        for attribute in self.attributes:
+            if "unit" not in attribute:
+                continue
+            self.parameters["definitions"][attribute] = self.attributes[attribute]
+
+
+class PlutoVtkDataset(VtkMixin, StaticPlutoDataset):
```
neutrinoceros, line 1075: this file is becoming a mess, I get it, but reorganizing the code makes `git blame` less useful so I tend to avoid moving classes around just for the sake of organizing. It should also be avoided in a PR that doesn't litteraly anything else because it inflates the diff and makes reviewing harder than it needs to be.

dutta-alankar:
> Sorry this took me so long to get back to you. I pushed a couple simplifications myself and added some additional comments and suggestions. I think we need to try and deduplicate as much as possible before I can merge.

Yes sure. I agree with you. Let me go through this and get back to you.

dutta-alankar:
@neutrinoceros A quick update. I need a few more days to get through everything you suggested and address them as I'm traveling right now. Sorry for the delay.

neutrinoceros:
Thanks, but no rush !

Issue #356, opened 2025-01-16
Trouble with reading h5 filename


RitaliG:
Getting an error 
"""File "h5py/h5o.pyx", line 190, in h5py.h5o.open
KeyError: 'Unable to open object (component not found)'" """
while trying to do a projection plot with the yt environment suddenly. 

To verify the timestep, I gave a print statement for the variable "entry" after line 127 in lib/python3.10/site-packages/yt_idefix/io.py. It returns 2025, instead of the timestep I was trying to access. There is a 2025 in the file path ("turb-sims-jan2025/test-wind/"). Is there a workaround for this?

dutta-alankar:
Got it! This has been an edge case which was overlooked indeed. I guess we didn't expect the directory structure to contain any 4 digit number. Instead we must look for 4 digit number at the end. 
I, therefore, propose the following regex in `io.py` line 112. See the [PR](https://github.com/neutrinoceros/yt_idefix/pull/357).

neutrinoceros:
Thanks @RitaliG  for reporting and @dutta-alankar for proposing a solution so promptly !
I'm not confident I have enough details yet to understand what the problem is or what the solution should be, but let's discuss it over at #357.

PR #357, opened 2025-01-16
Regex fix in io.py

dutta-alankar:
Simple regex fix for the issue [https://github.com/neutrinoceros/yt_idefix/issues/356](https://github.com/neutrinoceros/yt_idefix/issues/356)

### DIFF
--- a/src/yt_idefix/io.py
+++ b/src/yt_idefix/io.py
@@ -109,7 +109,7 @@ class PlutoXdmfIOHandler(BaseIOHandler):
             2 4.998045e+00 3.400969e-03 1458 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
             3 7.497932e+00 3.386245e-03 2186 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
         """
-        if (match := re.search(r"\d{4}", self.ds.filename)) is not None:
+        if (match := re.search(r"\d{4}(?!.*\d{4})", self.ds.filename)) is not None:
             entry = int(match.group())
         else:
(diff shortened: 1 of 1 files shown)


src/yt_idefix/io.py
```diff
@@ -109,7 +109,7 @@ def _read_fluid_selection(self, chunks, selector, fields, size):
             2 4.998045e+00 3.400969e-03 1458 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
             3 7.497932e+00 3.386245e-03 2186 single_file little rho vx1 vx2 vx3 prs tr1 tr2 tr3 Temp ndens PbykB mach
         """
-        if (match := re.search(r"\d{4}", self.ds.filename)) is not None:
+        if (match := re.search(r"\d{4}(?!.*\d{4})", self.ds.filename)) is not None:
```
neutrinoceros, line 112: I was able to find documentation for `?!` (negative look-ahead), but I never used this. I'll ask a couple naive questions to make sure I understand how this works.
Surrounding `()` makes it a non-capturing group, is that right ?
So, the content of `match` is unchanged when the search *does* succeed (good), but we now reject file names that have more than one substring matching `\d{4}` ?

If the goal is to only match the file *name* (excluding the directory), a simpler fix would be
```suggestion
        if (match := re.search(r"\d{4}", self.ds.basename)) is not None:
```

dutta-alankar: @neutrinoceros You are right. I wanted to just make this take the last match with 4 digits as it contains the filename. But I think what you are suggesting is also pretty good and will work.

neutrinoceros: Ok, since this is a pretty different approach I'll open a separate PR then.

PR #358, opened 2025-01-16
BUG: fix incorrect regexp subject (only match basenames, not entire paths)

### LINKED ISSUE #356: Trouble with reading h5 filename

neutrinoceros:
Close #356
Close #357 (which this is an alternative to)


neutrinoceros:
@RitaliG @dutta-alankar, could you confirm that this resolves the issue ?

RitaliG:
I checked, and this fixes the problem.
Many thanks @neutrinoceros and @dutta-alankar for the quick help.

neutrinoceros:
Awesome, I'll ship it as yt_idefix 2.4.1 right now then. Thank you !

PR #368, opened 2025-02-11
BUG: fix reading native grid coordinates for idefix 2.2 vtk files

Anto6453:
Fix `read_grid_coordinates()` method so that it returns cell edges when native coordinates are in the metadata.


neutrinoceros (approved):
Thanks a lot @Anto6453 !

PR #387, opened 2025-07-10
TST: ensure lock upgrades are compatible with Python version under test in devdeps job

neutrinoceros:


neutrinoceros:
I've found that this approach didn't work in other packages so closing without merge

PR #388, opened 2025-07-10
TST: re-implement devdeps locking+syncing with a more reliable strategy

neutrinoceros:


neutrinoceros:
this isn't ideal either because I seemingly cannot avoid building yt twice at the moment. It might get easier in the future if yt starts issuing stable ABI wheels

